PR #22, opened 2015-01-25
Fix regex to match text in all rows

btilm305:
Previously, the regex would only parse the css on the last line of the CSS Styles box.  Now, it parses every line.


braamgenis:
Thanks for the fix btilm305. Much appreciated!

PR #34, opened 2015-02-24
Add panel delta number to $panels_data

dimitrov-adrian:
Now it is easy to add CSS styling on widget id.


gregpriday:
Thanks again for all your contributions. I'm about to merge this in. Just made a minor change to the name of the index where the index is stored to cell_index, as in "index within the cell" - this made the most sense to me.

PR #37, opened 2015-02-24
Close widgets window with esc

dimitrov-adrian:
Allow user to close widget adding window with [Esc] keyboard key. Also adding keyboard listener handler function. Actually the closing is not in the very elegant form but it is good start :)


gregpriday:
Another very useful change. Thanks Adrian.

PR #40, opened 2015-02-25
Overwriting Page Builder Options

ziga-miklic:
I noticed that overwriting options using add_theme_support( 'siteorigin-panels') does not work. The above changes fixes this. I only replaced the order of $current_settings and $theme_settings, so that way the theme options are applied last.

I tested it on my theme and it works as intended.

### DIFF
--- a/inc/options.php
+++ b/inc/options.php
@@ -35,7 +35,7 @@ function siteorigin_panels_setting($key = ''){
 		if( !empty($theme_settings) ) $theme_settings = $theme_settings[0];
 		else $theme_settings = array();
 
-		$settings = wp_parse_args( $theme_settings, apply_filters( 'siteorigin_panels_settings_defaults', array(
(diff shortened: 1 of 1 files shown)


gregpriday:
Hi Žiga

Thanks for the PR. The add_theme_support setting values are actually a little confusing. As it stands, users expect that once they've set and saved a setting, that it shouldn't be changed. A good example of this is the bottom margin. Old versions of Page Builder used to give higher priority to add_theme_support values, but users got frustrated that they couldn't change the bottom margin.

Ideally, certain settings (like the bottom margin), should be stored as theme mods, allowing a theme to provide a theme specific default value, but still let the user change the value.

We're currently working on more advanced settings to Page Builder, so I'll see if it'd be possible to come up with a solution here.

braamgenis:
Closing old PR as stale.

PR #44, opened 2015-03-15
Current post id to style fields filters

dimitrov-adrian:
Allow siteorigin_panels_row_style_fields and siteorigin_panels_widget_style_fields to know from which post_id they are called.


gregpriday:
Very good work. I can see how this will open up a lot of options for developers. Github can't do an automatic merge, so I'll do a manual merge as soon as possible.

Or could you pull in the latest develop branch and resubmit the PR. The major change in develop is how we're handling super globals now. We're using the filter_input PHP function.

https://github.com/siteorigin/siteorigin-panels/pull/44/files#diff-9309031af65db361b95cbfd5ba250addR11

PR #45, opened 2015-03-16
Separate mesurement list

dimitrov-adrian:
I know there is no point of it at first glance but, actually in some cases alowing users to select in 'ex' or 'pt' will be overkill. Also with fn there is a way to uniform plugin enhancements.


gregpriday:
Any PR that adds a few more points for developers to integrate with are a great idea if you ask me. Thanks @dimitrov-adrian.

PR #52, opened 2015-03-22
Allow multi values

dimitrov-adrian:
Supports measurements values like '1 0' [em] wich will provide 1em 0em or so, if for field is set property multiple (example usage is paddings for widgets)

$fields['padding']['multiple'] = 1;

### DIFF
--- a/inc/styles.php
+++ b/inc/styles.php
@@ -315,9 +315,24 @@ function siteorigin_panels_sanitize_style_fields($section, $styles){
 				$return[$k] = !empty( $styles[$k] );
 				break;
 			case 'measurement' :
-				preg_match('/([0-9\.,]+)(.*)/', $styles[$k], $match);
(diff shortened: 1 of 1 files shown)


dimitrov-adrian:
Example usage in SiteOrigin_Panels_Default_Styling::widget_style_fields()

```
$fields['padding'] = array(
			'name' => __('Padding', 'siteorigin-panels'),
			'type' => 'measurement',
			'group' => 'layout',
			'description' => __('Padding around the entire widget.', 'siteorigin-panels'),
			'priority' => 7,
			'multiple' => TRUE
		);
```

It is safe to use in this case because it not mess with the styling negative paddings.

gregpriday:
Oh wow, this is an incredibly clever idea. I'll review this as soon as possible.

gregpriday:
Just noticed this was auto closed while I was cleaning up old branches. If you have a moment, could you resubmit this to the develop branch? If not, I'll try get it in.

PR #60, opened 2015-04-09
Allow panels_info to be passed to instance

aduth:
I assume there's some reasoning the `panels_info` has explicitly been unset from the instance, but I have a seemingly basic use case that is difficult to achieve given the current behavior. Using `get_less_variables`, I'd like to be able to pass the configured `font_color` to the LESS styles so that I can lighten or darken colors in the stylesheet based on the configured font color. The proposed changes make this possible by allowing me to access `$instance['panels_info']['style']['font_color']`. Otherwise, without explicitly calling to and iterating over the `panels_data` post meta, I can't imagine any way to use these configuration options.

Any suggestions for another approach I could be taking here?

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -947,12 +947,9 @@ function siteorigin_panels_render( $post_id = false, $enqueue_css = true, $panel
 			if( !empty($cell_style_wrapper) ) echo $cell_style_wrapper;
 
 			foreach ( $widgets as $pi => $widget_info ) {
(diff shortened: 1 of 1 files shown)


gregpriday:
This was actually behaviour that was only really necessary when we used an unnamespaced argument for `panels_info` (it used to just be `info`). As long as widgets are able to properly deal with only sometimes having panels_data available, it shouldn't be a problem having it in `$instance`.

PR #65, opened 2015-04-23
Allow themes and plugins interact with wpColorPicker

dimitrov-adrian:
Allow themes and plugins interact with wpColorPicker plugin settings. Useful in case of changing default color suggestions.


gregpriday:
This is an interesting looking change. How would a theme go about adding custom settings? Would it need to modify the soPanelsOptions global to add wpColorPickerOptions?

Another option would be to also add a filterable array for the wpColorPickerOptions in the `wp_localize_script` call. That might be a little easier to use.

dimitrov-adrian:
Actually most usable here is the *palettes* parameter. Other could be *mode*. Other parameter could break the the behavior of the color picker. But yeah, may be you are right, it will be more suitable for devs to use filterable array of *so-panels-admin* localization in siteorigin_panels_admin_enqueue_scripts().

PR #75, opened 2015-07-05
Prevent PHP warning when using additional field.

dimitrov-adrian:


gregpriday:
Nice little check there Adrian. Thanks!

PR #89, opened 2015-08-13
Remove Margin Bottom Unless Set

unetics:
Only adds inline margin bottom if set in page builder really helpful for custom css.


gregpriday:
I was actually just chatting to a friend about this the other day. The only problem I can see is that for existing users, all the bottom margins that they were expecting to be there by default will disappear.

So we'd need to make sure that users opt in to this behaviour first. Maybe setting a bottom margin of `0` for a row could remove the bottom margin CSS entirely.

unetics:
i will take a look and see if i can make a bit more friendly been working with this mod for a wile didn't realise it removed default margins (i always have set margin bottom to 0 in settings panel)

braamgenis:
Closing old PR as stale.

PR #92, opened 2015-08-19
Make JS_SUFFIX constant overrideable

MatthewEppelsheimer:
Perhaps a bit of an edge case, but this checks to see whether the constant `SITEORIGIN_PANELS_JS_SUFFIX` is already defined, before defining it. 

This will allow developers to use non-minified scripts in local environments for debugging, without needing to run the `develop` branch of the plugin. Instead, they can define the constant as a blank string in their `wp-config.php` file. 

I looked and don't see any way this will interfere with the `gulpfile.js` release build script, but it may be a good idea to check.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -12,7 +12,9 @@ Donate link: http://siteorigin.com/page-builder/#donate
 */
 
 define('SITEORIGIN_PANELS_VERSION', 'dev');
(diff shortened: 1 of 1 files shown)


gregpriday:
Yeah, this definitely makes sense. Thanks @MatthewEppelsheimer

PR #93, opened 2015-08-21
Send widget class when fetching styles via AJAX

leeps:
When fetching widget styles, include the widget class from the model.
This is needed if certain style attributes should only be available for
certain widgets, for example in the
siteorigin_panels_widget_style_fields
filter.


braamgenis:
Hi @leeps Sorry we never got around to looking at this, but due to the time elapsed I'm closing this PR as stale. Feel free to open a new issue for discussion.

Issue #85, opened 2015-08-01
PB not honouring the classname for the widget output


primozcigler:
Hi Greg,

A question: we noticed in our team that Page Builder is not honouring the `classname` if supplied to the [__construct function](https://developer.wordpress.org/reference/classes/wp_widget/__construct/) when defining a widget?

This code: https://github.com/siteorigin/siteorigin-panels/blob/develop/siteorigin-panels.php#L1134

It is always defining the HTML class name based on the `id_base` property, ignoring the optional classname.

I am wondering if there is any particular reason for that? If not, I would love to create a PR with that change.

(I am aware it would be possible to add that via filter `siteorigin_panels_widget_classes`, but I believe this should be fixed in the plugin itself.)

Thank you for the answer!

primozcigler:
Hi @gregpriday, I will really appreciate an answer :)

Thanks!

gregpriday:
Hi @primozcigler! Sorry for the late reply here - really bad of me. There's no good reason for Page Builder to be ignoring the classname attribute. This is just an oversight on my behalf.

If you submit a PR that fixes this, I'd be happy to merge it. I'll take a look at this myself if you can't find the time though.

primozcigler:
Hi Greg,

I am GLAD to hear that! :smile: 

Expect the PR later today.

PR #95, opened 2015-08-25
Add classname of the widget to HTML output if classname is defined

primozcigler:
As discussed in #85


gregpriday:
Much appreciated good sir :+1:

PR #100, opened 2015-09-03
Adding new hook for adding HTML before row

galethil:
New filter 'siteorigin_panels_before_row_inside_grid' allows themes or plugins to add HTML at the beggining of each row, but inside panel-grid DIV element. This is usefull for corner ribbons as well as for other design elements.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -927,6 +927,9 @@ function siteorigin_panels_render( $post_id = false, $enqueue_css = true, $panel
 			echo $name.'="'.esc_attr($value).'" ';
 		}
 		echo '>';
+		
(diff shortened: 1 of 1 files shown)


galethil:
@gregpriday could you please take a look at my pull request. It will be a big help for me. Thanks

braamgenis:
Hi @galethil 

Unfortunately we missed this PR and have made some big changes which would conflict with your change.

The good news is, we now have a filter `siteorigin_panels_before_content` which would allow you to add HTML before the row content. You can see it here: https://github.com/siteorigin/siteorigin-panels/blob/develop/siteorigin-panels.php#L974

Sorry for the misunderstanding and thanks for your contribution.

PR #105, opened 2015-10-07
Create siteorigin-panels-cs_CZ.po

chuckumi:
Czech translation 89%


gregpriday:
This is now handled mostly by Glotpress.

PR #111, opened 2015-10-21
Set 'additional' key on query_args

rjv:
The query args 'additional' key was not being properly set to an empty array if no value was given. This caused a PHP warning when DEBUG is enabled. This modification ensures that $query_args['additional'] always gets set.


gregpriday:
Thanks for the contribution @rjv. Everything looks good to me, but @braamgenis, can you check this out when you get back from your holiday. It's in your part of the code so I'll let you merge.

braamgenis:
Looks good! Thanks for this @rjv.

PR #116, opened 2015-11-26
Changed the #content-panels anchor to look for the correct span after…

richardtape:
… WP 4.4 switched from using a h3 to a h2. This now places the correct text in the anchor (rather than 'undefined'). Also makes it more generic by not over qualifying the selector so should WP change headings again upstream, this shouldn't happen. Fixes 115.


richardtape:
Wrong branch. Will fix.

Issue #129, opened 2015-12-09
Row style option to change collapse order.


gregpriday:
Right now, rows always collapse so that the column on the right goes on top and the column on the left goes at the bottom. People have requested an option to be able to change this ordering on a row by row basis.

gregpriday:
Have this working here in a feature branch.

https://github.com/siteorigin/siteorigin-panels/tree/feature/reverse-collapse-rows

PR #133, opened 2015-12-27
Feature/reverse collapse rows

gregpriday:
Resolves GH-129. Also makes RTL support more efficient by changing where the order is reversed.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -732,11 +732,14 @@ function siteorigin_panels_generate_css($post_id, $panels_data = false){
 			));
 		}
 
+
+		$collapse_order = !empty( $grid['style']['collapse_order'] ) ? $grid['style']['collapse_order'] : ( !is_rtl() ? 'left-top' : 'right-top' );
+
 		if ( $cell_count > 1 ) {
-			$css->add_cell_css($post_id, $gi, false, '', array(
+			$css->add_cell_css( $post_id, $gi, false, '', array(
 				// Float right for RTL
-				'float' => !is_rtl() ? 'left' : 'right'
-			));
+				'float' => $collapse_order == 'left-top' ? 'left' : 'right'
+			) );
 		}
 
 		if ( $settings['responsive'] ) {
@@ -747,7 +750,7 @@ function siteorigin_panels_generate_css($post_id, $panels_data = false){
 			), $panels_mobile_width);
 
 			for ( $i = 0; $i < $cell_count; $i++ ) {
-				if ( $i != $cell_count - 1 ) {
+				if ( ( $collapse_order == 'left-top' && $i != $cell_count - 1 ) || $collapse_order == 'right-top' &&  $i != 0 ) {
 					$css->add_cell_css($post_id, $gi, $i, '', array(
 						'margin-bottom' => $panels_margin_bottom . 'px',
 					), $panels_mobile_width);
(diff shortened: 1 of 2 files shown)


siteorigin-panels.php
```diff
@@ -732,11 +732,14 @@ function siteorigin_panels_generate_css($post_id, $panels_data = false){
 			));
 		}
 
+
+		$collapse_order = !empty( $grid['style']['collapse_order'] ) ? $grid['style']['collapse_order'] : ( !is_rtl() ? 'left-top' : 'right-top' );
```
braamgenis: Shouldn't `is_rtl()` still make it switch even if the collapse order was set by the user?

siteorigin-panels.php
```diff
@@ -925,7 +928,8 @@ function siteorigin_panels_render( $post_id = false, $enqueue_css = true, $panel
 		}
 	}
 
-	if( is_rtl() ) $panels_data = siteorigin_panels_make_rtl( $panels_data );
+	// A filter to change the panels_data right before the content is rendered.
+	$panels_data = apply_filters( 'siteorigin_panels_data_before_render', $panels_data, $post_id );
```
braamgenis: Shouldn't this filter now be added somewhere to do whatever `siteorigin_panels_make_rtl` was doing?

siteorigin-panels.php
```diff
@@ -1008,6 +1012,12 @@ function siteorigin_panels_render( $post_id = false, $enqueue_css = true, $panel
 		$row_style_wrapper = siteorigin_panels_start_style_wrapper( 'row', $style_attributes, !empty($panels_data['grids'][$gi]['style']) ? $panels_data['grids'][$gi]['style'] : array() );
 		if( !empty($row_style_wrapper) ) echo $row_style_wrapper;
 
+		$collapse_order = !empty( $panels_data['grids'][$gi]['style']['collapse_order'] ) ? $panels_data['grids'][$gi]['style']['collapse_order'] : ( !is_rtl() ? 'left-top' : 'right-top' );
```
braamgenis: Same question here as previously about `is_rtl` still causing it to toggle if the option set by a user.

siteorigin-panels.php
```diff
@@ -1365,35 +1375,6 @@ function siteorigin_panels_render_form($widget, $instance = array(), $raw = fals
 }
 
 /**
- * This takes existing Page Builder data and makes it RTL by reversing the content
- */
-function siteorigin_panels_make_rtl($panels_data){
```
braamgenis: Ah, ok, I see you removed `siteorigin_panels_make_rtl`. so you can ignore the previous comment about the added filter.

siteorigin-panels.php
gregpriday: I think once a user has manually set the collapse order, we should obey that fairly strictly. It's only when they've selected _Default_ that we'd want to use `is_rtl()` to work out the best option.

siteorigin-panels.php
gregpriday: That was the original intention, but the way we're reversing the cells now, on the processed version of the data, is a lot more effective.

https://github.com/siteorigin/siteorigin-panels/pull/133/files#diff-72e0505cdd6cb4ce7175f772c9c68bafR1018

It does mean this new filter isn't all that important though. We could remove it.

siteorigin-panels.php
braamgenis: Ok, I think it's fair enough either way. Just wondered about the case where a user might want the opposite of the deault `is_rtl()` behaviour.

siteorigin-panels.php
braamgenis: Cool. Then I don't think it will hurt to leave it in, but up to you.

braamgenis:
Cool, all good. Will leave to you if you want to make any changes before merging.

braamgenis:
Seem to be some conflicts, so I'll leave to you to merge.

Issue #54, opened 2015-03-24
Last row margin


alexilin83:
Hi.
Last row margin is missing, no matter what's written in the settings.

braamgenis:
Hi @alexilin83 
Apologies. This was being done intentionally to avoid accidentally changing the containing theme's layout.

PR #134, opened 2016-01-06
Last row margin #54

KevinTaron:
The fix for issue "Last row margin #54"
The last grid was ignored. Don't know why.


braamgenis:
Hi @KevinTaron 

Thanks very much for your input, but this was being done intentionally to avoid changing the containing theme's layout.

Issue #137, opened 2016-01-07
If i disable "Add Widget Class"  from setting Live Editor not works.


amanjaswalia:
If i disable "Add Widget Class" Live Editor Not works. I think Live Editor works with .widget class on page builder.

gregpriday:
Thanks for letting us know, I just confirmed the issue on my side.

PR #138, opened 2016-01-08
For GH-137: live editor not working when `.widget` class not used.

braamgenis:
Use `.so-panel` class to identify widget panels in live editor.


gregpriday:
Looks good

braamgenis:
Resolve issue GH-137

PR #143, opened 2016-01-19
Fix for compatibility issues with Yoast SEO.

braamgenis:
This immediately copies PB content across to the default TinyMCE editor if it's available and if not, retries 3 times with small delays. Additionally it fires a 'change' event when the content is changed, which should trigger Yoast SEO's handling of the content.


gregpriday:
Do you want to push this to the new `gulp-browserify` branch instead? I'll be merging it into `develop` soon.

https://github.com/siteorigin/siteorigin-panels/tree/feature/gulp-browserify

PR #144, opened 2016-01-20
Tablet responsive break point setting

braamgenis:
Added tablet responsive setting, which defaults to 1024px and simply sets column widths to 50%.


gregpriday:
Looks golden

PR #146, opened 2016-01-21
Feature/custom row ids

gregpriday:
This allows users to enter custom row IDs. Had to add some code to `siteorigin_panels_process_panels_data` to make sure the IDs are numeric and at least unique within a given page.


braamgenis:
It's a bit hard to follow this code: 3cad425b2bf62fe71fd5304d73a6bd5faa8d08ce so would be nice to have an extra comment or two, but up to you. Otherwise good to merge when you're ready. :ok_hand:

live-mesh:
Sorry I am not able to go through the change fully to obtain a better understanding of what this involves. I am just wondering - what's wrong with the below code and leaving it to the user to ensure that id is unique (like VC does) and does not disturb the existing auto-generated unique row id used for other purposes by the page builder. - 

`
...
        add_filter('siteorigin_panels_row_style_fields', array($this, 'row_style_fields' ) );

        add_filter('siteorigin_panels_row_style_attributes', array($this, 'row_style_attributes' ), 10, 2);

}

    function row_style_fields($fields) {
        // Add the attribute fields

        $fields['row_id'] = array(
            'name' => __('Row Id', 'siteorigin-panels'),
            'type' => 'text',
            'group' => 'attributes',
            'description' => __('An ID for the row', 'siteorigin-panels'),
            'priority' => 4,
        );

        return $fields;
    }

    function row_style_attributes( $attributes, $args ) {

        // Do not set id if already set. Hope to get rid of this in future when page builder has this function
        if( !empty( $args['row_id'] ) && empty($attributes['id'])) {
            $attributes['id'] = $args['row_id'];
        }

        return $attributes;
    }`

PR #148, opened 2016-01-21
add hook to inject HTML before and after cell loop; fix #147

### LINKED ISSUE #147: Add hook to inject HTML before and after cell loop
I am proposing here to add hook to add HTML before and after cell loop while rendering panels in `siteorigin_panels_render` function.

There are hooks to add HTML content before and after row. But I want to wrap all cells with a div to apply extra styling. Adding those hooks will be great. I am submitting patch later.

Thanks

ernilambar:
There are hooks to add HTML content before and after row. But I want to wrap all cells with a div to apply extra styling. Adding those hooks will be great. I am submitting patch later.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -1027,6 +1027,8 @@ function siteorigin_panels_render( $post_id = false, $enqueue_css = true, $panel
 			$cells = array_reverse( $cells, true );
 		}
 
+		echo apply_filters( 'siteorigin_panels_before_cell_loop', '', $panels_data['grids'][$gi], $grid_attributes );
+
 		foreach ( $cells as $ci => $widgets ) {
 			// Themes can add their own styles to cells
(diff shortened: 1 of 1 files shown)


gregpriday:
Thanks for the contribution @ernilambar! I'll take a proper look at this when I have some fresher eyes. I know it's a fairly straight forward PR, but for things like adding hooks and filters, we want to make sure that they're added in a very consistent way. Things like naming conventions, arguments and even the order of the arguments are all important.

ernilambar:
Make sense. We should definitely follow a standard for adding action and filter. Will keep on eye on this ticket and hope we will be able to make decision about it. Now I am using very hacky way by using custom javascript for wrapping cell elements.

brettsmason:
I would love to see this added - this would solve a current headache I have very easily :+1:
Great plugin by the way! :smile:

gregpriday:
We actually added this with the new renderer. It was part of a broader rewrite of the renderer. It'll be available in the Page Builder 2.5 update.

PR #151, opened 2016-01-25
Adds possibility to have open groups on widget/row open

derpoho:
Was a need for us, makes handling the options in the sidebar to be faster accessed.

When you add a custom group via filter you can add 'open' => true to the array, and the group starts openend in the sidebar.

Hope you implement this..;)


gregpriday:
That would be a very nice addition. Thanks for the contribution! I just
need to double check this for consistency with how we're handling
expanded/open options in other areas, like the Widgets Bundle.

clubdesign:
I did check widgets and row options, these work fine. Didn`t check other areas yet.

braamgenis:
Closing old PR as stale. Feel free to open an issue if you'd like to discuss.

Issue #78, opened 2015-07-21
Append prebuilt layouts instead of replacing the content with them


h8every1:
It would be nice if it was possible to append prebuilt layouts instead of replacing the whole content with them.

I have some handy snippets in my Prebuilts. For example two column layout with WYSIWYG-editor in one column and image widget in the second. They have some CSS classes and styling applied.

If I already have some content, I can't insert this snippet, because I will lose all my content.

Maybe I'm missing something at is laready possible to achieve what I want?

cmwwebfx:
So what i understand from this comment you would like to save row layouts for reuse on the site?

h8every1:
Yes. And most important thing is that i want to be able to use several row
layouts in a post.

вторник, 21 июля 2015 г. пользователь Ciaran Whelan написал:

brentjett:
+1  I don't currently use prebuilts because of this behavior. Would love to construct pages with pre defined rows.

gregpriday:
I worked on this briefly a little while ago, but I'll give it another shot.

danpilat:
+1 - we would love this!

Sekoul:
+1 - ditto, would be great. Thank you!

PR #162, opened 2016-02-18
Allow importing layouts before/after existing layout.

braamgenis:
To resolve GH-78.

Opening the PR for review, but we can still make any changes if we need to.


gregpriday:
Good point on not making descructive behaviour default. Hard one. So the 2 arguments for using replace as defaults are

1) That's how we did it before. Always need to put some thought into changing previous default behaviour. 
2) A common use case of the prebuilt layout would be a user trying out a few different layouts. So importing, checking, then trying a different one.

Although I'm not sure which is more frustrating. Accidently appending a layout, or accidently destroying a layout. My guess is that most of the time a user importing a layout will want to use that layout as their new layout. Appending and prepending are a slightly more niche use case.

We can think about it for a bit.

gregpriday:
Another option for the dropdown is to have it blank to start. This will force the user to choose if they want to replace or append before clicking the insert button. Ideally, this should be stored so that Page Builder remembers what a user chose and selects that whenever the dialog is displayed in the future.

Also, for the insert button. At the moment, if the user double clicks, then they don't really see the confirmation. Ideally, we should ignore double clicks. Maybe adding in a short timeout before accepting the second click would work. Will need to play around, but my guess is something around 100ms.

gregpriday:
Rad, everything looking good here. No harm in merging and doing any minor interaction tweaks before the next release.

PR #163, opened 2016-02-18
Ensure PB content is cleared when switching back to visual editor.

braamgenis:
@gregpriday Please take a look. I'm not entirely sure why we were skipping setting PB content on the visual editor when it was empty, other than optimization. This was leaving behind PB content in the editor when the expectation seems to be that it should have been cleared.

### DIFF
--- a/js/siteorigin-panels.js
+++ b/js/siteorigin-panels.js
@@ -2925,7 +2925,7 @@ module.exports = Backbone.View.extend( {
         });
 
         // When the data changes in the model, store it in the field
-        this.model.on('change:data', this.storeModelData, this);
+        this.model.on('change:data load_panels_data', this.storeModelData, this);
 
         // Handle a content change
         this.on('content_change', this.handleContentChange, this);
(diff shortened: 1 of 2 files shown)


gregpriday:
Yeah, I can't actually think of a good reason. No harm in removing that check.

What's this change do though? https://github.com/siteorigin/siteorigin-panels/pull/163/files#diff-83dd5f80393228689a7367e94915ca16R2928

braamgenis:
Yeah, it's quite involved. I'll try explain:
1. The 'Switch to Editor' button is clicked and user confirms in the alert that they'd like to clear PB content.
2. The PB model is cleared by calling [`loadPanelsData(false)`](https://github.com/siteorigin/siteorigin-panels/blob/gh-114-not-clearing-content/js/siteorigin-panels/view/builder.js#L179) which triggers the `load_panels_data` event.
3. The visual editor content is updated when the [`content_change`](https://github.com/siteorigin/siteorigin-panels/blob/gh-114-not-clearing-content/js/siteorigin-panels/view/builder.js#L319) event is triggered by the `storeModelData` method, which is bound to the model's `change:data` event.

So to ensure that the visual editor content is updated consistently and reflects what is in the model, I either had to trigger a `change:data` event in `loadPanelsData` or bind `storeModelData` to `load_panels_data` event. I went with the latter option as I think it was a simpler change than have `loadPanelsData` trigger two events or changing the `load_panels_data` event to rather be a `change:data` event too.

gregpriday:
Awesome. Sounds good. Had to think hard to figure that one out.

Issue #122, opened 2015-11-30
Appearance > Home Page needs revision capabilities 


Misplon:
SiteOrigin themes that add the Appearance > Home Page item are missing revisions that would be found if you edit the page via Pages > Home.

gregpriday:
When that home interface is saved, we need to hack it to create a new revision for the page that it's using as the home page.

WordPress uses a custom post type as post revisions. It should be fairly easy to create a fake revision. We'd just need to add post meta `panels_data` to that and set the `post_content`. Once you're done, making changes in this appearance > Home Page interface should add revisions when you go to Edit as Page.

PR #169, opened 2016-02-24
Custom Home Page revisions

braamgenis:
To resolve GH-122. Include post_content when saving changes to Custom Home page to allow saving revisions.


js/siteorigin-panels.js
```diff
@@ -3469,13 +3469,23 @@ module.exports = Backbone.View.extend( {
         // Make sure we actually need to copy content.
         if( panelsOptions.copy_content && this.attachedToEditor && this.$el.is(':visible')) {
```
gregpriday: Does the Home Page interface attach itself to the editor? Does it pass `this.attachedToEditor`?

braamgenis: Indeed it does. ¯\_(ツ)_/¯ Which is a tad whack, isn't it.

This part is somewhat hacky and probably needs a slight rethink.

gregpriday: Hrmm, that's whack. Yeah, we shouldn't rely on passing this check. We might fix that variable in the future and then break this particular content copy mechanism.

braamgenis: ACK. Will fix this up.

gregpriday: If you spot the issue, feel free to fix up `attachedToEditor` while you're at it ;)

braamgenis:
@gregpriday Made some changes to try fix up `attachToEditor` and not use that check to update content for the custom home page builder.

gregpriday:
How does this work with Layout Widgets now? That's a case where we have a Builder view, that shouldn't be copying content.

There's also the case of other plugins using a Builder view in different ways. For example, our still unreleased [Post Type Builder](https://github.com/siteorigin/so-cpt-builder).

So my guess here is that we need to modify the attachToEditor function so that it also handles being attached to a hidden text field. Then for the home page interface, we could just attach the builder to the new hidden text field.

braamgenis:
Ok, so it should only be trying to copy across content for main PB and Custom Home pages.

gregpriday:
I need to change my IDE indenting settings so that we indent with tabs like the WordPress coding guidelines say. See you're using tabs, but I've been using spaces. My bad.

Will double check this PR as soon as I can.

braamgenis:
You're using IntelliJ/PHPStorm, right? We have an `.editorconfig` file so you should just be able to enable [EditorConfig](http://editorconfig.org/) support and all will be well. :smile: Open Settings > Editor > Code Style and there should be a checkbox at the bottom.

js/siteorigin-panels/view/builder.js
```diff
@@ -549,13 +563,23 @@ module.exports = Backbone.View.extend( {
         // Make sure we actually need to copy content.
         if( panelsOptions.copy_content && this.attachedToEditor && this.$el.is(':visible')) {
 
+			var postId;
+			if(this.builderType === this.builderTypes.EDITOR_ATTACHED) {
+				postId = $('#post_ID').val();
+			} else if(this.builderType === this.builderTypes.CUSTOM_HOME_PAGE) {
+				postId = $('#panels-home-page').data('postId');
+			} else {
+				// This shouldn't happen.
```
gregpriday: This actually can happen. I'm not sure if the `builderType` I added is the same as the one you added, but the reason I originally did this was to pass the builder type along with the request for style fields.

The builder type can be set by anything that wants to use a builder view. For example the Post Type Builder sets the `data-type` field [here](https://github.com/siteorigin/so-cpt-builder/blob/develop/tpl/admin-build.php#L5), which is used by Page Builder [here](https://github.com/siteorigin/siteorigin-panels/blob/develop/js/siteorigin-panels/main.js#L92).

Just checking that this PR takes this into account.

PR #170, opened 2016-02-26
Dropdown button for prebuilt dialog options.

braamgenis:
Basic dialog button handling for elements in toolbar with .so-toolbar-button class.

### DIFF
--- a/css/admin.css
+++ b/css/admin.css
@@ -1582,17 +1582,15 @@
   width: 125px;
   height: 28px;
 }
-.so-panels-dialog .so-dropdown-wrapper .so-dropdown-links-wrapper,
-.so-panels-dialog .so-dropdown-wrapper input[type="button"][disabled].button-primary:hover + .so-dropdown-links-wrapper {
-  display: none;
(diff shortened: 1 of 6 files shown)


gregpriday:
It looks like this works the first time, but if you try insert a layout for the second time on a given page, it doesn't work.

Also, ideally we'd have a sort of confirmation when the user clicks on the main button.

Reckon this is a big interface improvement though.

braamgenis:
@gregpriday Cool, I fixed the issue of inserting only working once. For the main button, I wonder if it should trigger an action at all? Just a thought, but shouldn't it rather just do nothing and require the user to select one of the dropdown (or is it dropup? :stuck_out_tongue: ) options?

gregpriday:
But buttons love to be pushed! Looking at how Bootstrap treats these it seems like they use the click action to show the dropdown/riseup. Just hovering does nothing. So maybe this is the behaviour we want to use then.

braamgenis:
Ok, it's working on click now. Hopefully that does the trick.

gregpriday:
Yeah, think this is the one.

Issue #132, opened 2015-12-19
Hide dialog sidebar when there's nothing to display


gregpriday:
If a user filters out all the row/widget styling options, Page Builder still shows the empty sidebar. Ideally this should be hidden.

`add_filter( 'siteorigin_panels_widget_style_fields', '__return_false', 11 );`

<img width="1648" alt="njutm94a6g" src="https://cloud.githubusercontent.com/assets/1126859/11913034/874ab4dc-a65f-11e5-9d23-8d4451b01941.png">

PR #176, opened 2016-03-03
Hide empty right sidebar in row/widget dialogs.

braamgenis:
This happens when style fields have been filtered out.
To resolve GH-132.


braamgenis:
This shows the loading spinner and then hides the sidebar when the AJAX call returns an empty response. Alternatively, might be nicer to hide the sidebars by default and then only show them if the response isn't empty.

gregpriday:
This solution is decent for now. It's a fairly rare use case.

Issue #177, opened 2016-03-03
Allow setting a bottom margin on the last row


jordif:
I understand the last row should not get a bottom margin by default. However, if a user wants to add a margin using the Row Styles dialog, I think they should be allowed to. Right now if you enter a value for the last row, it won't work.

I have solved this by editing the following line:

https://github.com/siteorigin/siteorigin-panels/blob/develop/siteorigin-panels.php#L723

And changing it to:

    if($gi != count($panels_data['grids'])-1 || !empty($grid['style']['bottom_margin'])){

Do you think it would make sense adding this in the next version of the Page Builder?

Thanks!

Jordi

braamgenis:
Fixed in GH-178

PR #178, opened 2016-03-04
Added last row margin features

KevinTaron:
#177 I've added a checkbox in settings so a user can activate "margin in last row" for every site.
If the user only want a margin in last row he can overwrite it with the row styles like @jordif did.

(Need that feature on some of my sites. At the moment i'm adding a empty row on the bottom to each page for this feature)

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -693,6 +693,7 @@ function siteorigin_panels_generate_css($post_id, $panels_data = false){
 	$panels_tablet_width = $settings['tablet-width'];
 	$panels_mobile_width = $settings['mobile-width'];
 	$panels_margin_bottom = $settings['margin-bottom'];
+	$panels_margin_bottom_last_row = $settings['margin-bottom-last-row'];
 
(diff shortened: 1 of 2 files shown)


siteorigin-panels.php
```diff
@@ -719,12 +720,19 @@ function siteorigin_panels_generate_css($post_id, $panels_data = false){
 			}
 		}
 
-		// Add the bottom margin to any grids that aren't the last
-		if($gi != count($panels_data['grids'])-1){
+		if($panels_margin_bottom_last_row) {
```
braamgenis: You could accomplish the same thing by just adding ` || !empty( $panels_margin_bottom_last_row )` to the original conditional expression. So in future, if anything needs to change about how we're calling `$css->add_row_css` we only have to change it in one place.

braamgenis:
Hi @KevinTaron Thanks for your contribution! I just have one suggestion about your code. Thanks again! :+1:

KevinTaron:
Hi @braamgenis you are totaly right. Don't know why did that this way ... Fixed ;)

braamgenis:
Great, merging!

PR #183, opened 2016-03-09
Hotfix WordPress 4 5 compatibility

braamgenis:


js/siteorigin-panels/model/widget.js
```diff
@@ -96,8 +96,9 @@ module.exports = Backbone.Model.extend( {
         this.set( 'values', values, {silent: true} );
 
         if( hasChanged ) {
-            // We'll trigger our own change events
-            this.trigger('change');
+            // We'll trigger our own change events.
+			// NB: Must include the model being changed (i.e. `this`) as a workaround for a bug in Backbone 1.2.3
+            this.trigger('change', this);
             this.trigger('change:values');
```
gregpriday: For consistency, doesn't this need a `this` too?

js/siteorigin-panels/collection/cells.js
```diff
@@ -1,7 +1,7 @@
 var panels = window.panels;
 
 module.exports = Backbone.Collection.extend( {
-    model: panels.cell,
+    model: panels.model.cell,
```
gregpriday: No idea how I missed this. Or how it was still working properly...

js/siteorigin-panels/model/widget.js
braamgenis: No, if anything I'd say it should be passing `values`, because that's what's being changed. I think it's fine, seeing as we're not doing anything with the changed `model` or `values`.

PR #190, opened 2016-03-19
New Live Editor

gregpriday:
Several improvements to the live editor. Most importantly, we now have a complete builder interface in the right column and there are a few improvements to how we reload the preview to make it feel more responsive.

### DIFF
--- a/js/siteorigin-panels/dialog/widget.js
+++ b/js/siteorigin-panels/dialog/widget.js
@@ -156,12 +156,12 @@ module.exports = panels.view.dialog.extend( {
         if( !this.model.get('missing') ) {
             // Only get the values for non missing widgets.
             var values = this.getFormValues();
-            if (typeof values.widgets === 'undefined') {
+            if ( typeof values.widgets === 'undefined' ) {
                 values = {};
             }
             else {
                 values = values.widgets;
-                values = values[Object.keys(values)[0]];
+                values = values[ Object.keys(values)[0] ];
             }
 
             this.model.setValues(values);
--- a/js/siteorigin-panels/dialog/widget.js
+++ b/js/siteorigin-panels/dialog/widget.js
@@ -115,12 +115,12 @@ module.exports = panels.view.dialog.extend( {
      */
     loadForm: function(){
         // don't load the form if this dialog hasn't been rendered yet
-        if( !this.$el.find('> *').length ) {
+        if( !this.$('> *').length ) {
             return;
         }
 
         var thisView = this;
-        this.$el.find('.so-content').addClass('so-panels-loading');
+        this.$('.so-content').addClass('so-panels-loading');
 
         var data = {
             'action' : 'so_panels_widget_form',
@@ -137,7 +137,7 @@ module.exports = panels.view.dialog.extend( {
                 var html = result.replace( /{\$id}/g, thisView.model.cid );
 
                 // Load this content into the form
-                thisView.$el.find('.so-content')
+                thisView.$('.so-content')
                     .removeClass('so-panels-loading')
                     .html(html);
 
@@ -145,7 +145,7 @@ module.exports = panels.view.dialog.extend( {
                 thisView.trigger('form_loaded', thisView);
 
                 // For legacy compatibility, trigger a panelsopen event
-                thisView.$el.find('.panel-dialog').trigger('panelsopen');
+                thisView.$('.panel-dialog').trigger('panelsopen');
 
                 // If the main dialog is closed from this point on, save the widget content
                 thisView.on('close_dialog', thisView.saveWidget, thisView);
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -231,7 +231,7 @@ module.exports = Backbone.View.extend( {
 	            toolbar.css('top', newTop);
             }
         };
-	    
+
 	    this.on('builder_resize', stickToolbar, this );
         $( document ).scroll( stickToolbar );
         stickToolbar();
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -211,8 +211,6 @@ module.exports = Backbone.View.extend( {
                 newTop += $('#wpadminbar').outerHeight();
             }
 
-	        console.log(newTop);
-
 	        if( newTop < 0 ) {
 		        toolbar.css('top', 0);
 		        return false;
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -570,11 +570,6 @@ module.exports = Backbone.View.extend( {
                 }.bind(this)
             );
         }
-
-        if( this.liveEditor !== false ) {
-            // Refresh the content of the builder
-            this.liveEditor.refreshPreview();
-        }
     },
 
     /**
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -487,10 +487,6 @@ module.exports = Backbone.View.extend( {
      * @returns {panels.view.builder}
      */
     addLiveEditor: function(postId){
-        if( typeof panels.view.liveEditor === 'undefined' ) {
-            return this;
-        }
-
         // Create the live editor and set the builder to this.
         this.liveEditor = new panels.view.liveEditor( { builder: this } );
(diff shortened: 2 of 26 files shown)


tpl/js-templates.php
```diff
@@ -5,66 +5,70 @@
...
-
-		<a href="#" class="so-tool-button so-prebuilt-add">
-			<span class="so-panels-icon so-panels-icon-cubes"></span>
-			<span class="so-button-text"><?php esc_attr_e('Prebuilt', 'siteorigin-panels') ?></span>
-		</a>
-
-		<?php if( !empty($post) ) : ?>
+			<a href="#" class="so-tool-button so-widget-add">
+				<span class="so-panels-icon so-panels-icon-plus"></span>
+				<span class="so-button-text"><?php esc_attr_e('Add Widget', 'siteorigin-panels') ?></span>
```
braamgenis: I realize it's not something you changed, but just pointing it out while I'm here. It's a small thing, but couldn't all these be using `_e(...`? (and similarly for 'Add Row', annd 'Prebuilt' below)

js/siteorigin-panels/view/live-editor.js
```diff
@@ -19,7 +20,7 @@ module.exports = Backbone.View.extend( {
      * Render the live editor
      */
     render: function(){
-        this.$el.html( this.template() );
+        this.setElement( this.template() );
```
braamgenis: Is this to prevent Backbone from wrapping the template in a `div` element? Should it happen every time view's `render` method is called? If not then maybe it should go in the `initialize` method so that it's only called once for the view.

js/siteorigin-panels/dialog/widget.js
```diff
@@ -156,12 +156,12 @@ module.exports = panels.view.dialog.extend( {
         if( !this.model.get('missing') ) {
             // Only get the values for non missing widgets.
             var values = this.getFormValues();
-            if (typeof values.widgets === 'undefined') {
+            if ( typeof values.widgets === 'undefined' ) {
```
braamgenis: Not a problem, just a question. Are we following [jQuery style guide](https://contribute.jquery.org/style-guide/js/) conventions? We should probably decide which convention we're using and make a note of it somewhere in our docs or readme's.

js/siteorigin-panels/view/builder.js
```diff
@@ -475,10 +476,8 @@ module.exports = Backbone.View.extend( {
         }
 
         // Create the live editor and set the builder to this.
-        this.liveEditor = new panels.view.liveEditor();
-        this.liveEditor.setPostId(postId);
-
-        this.liveEditor.builder = this;
+        this.liveEditor = new panels.view.liveEditor( { builder: this } );
+        this.liveEditor.setPostId( postId );
```
braamgenis: I don't see the `setPostId` being called anywhere after this initial call so why not just have `postId` as an option in the constructor and set it in the `initialize` method?
E.g. `this.liveEditor = new panels.view.liveEditor( { builder: this, postId:postId } );`

js/siteorigin-panels/view/live-editor.js
```diff
@@ -32,7 +32,47 @@ module.exports = Backbone.View.extend( {
 		    // Scroll to the correct position
 		    thisView.$el.find( '.so-preview iframe' ).contents().scrollTop( thisView.previewScrollTop );
 
+		    var $$ = $(this ),
+			    ifc = $$.contents();
```
braamgenis: I wouldn't mind a nicer variable name. :stuck_out_tongue:  Perhaps `$iframeContainer`?

tpl/js-templates.php
```diff
@@ -519,7 +519,7 @@
 		</form>
 
 		<div class="so-sidebar-tools">
-			<a href="#" class="live-editor-close" title="<?php esc_attr_e('Close Live Editor', 'siteorigin-panels') ?>"></a>
+			<button href="#" class="live-editor-close button-primary"><?php esc_attr_e('Done', 'siteorigin-panels') ?></button>
```
braamgenis: Same as before. Necessary to have `esc_attr_e` and not just `_e`? Otherwise, I guess it should be `esc_html_e`?

braamgenis:
Looking good. Couple of comments, but nothing major. So good to merge, whenever you want. :metal: 

A bit of testing also revealed the following:
- CPT plugin doesn't load PB interface. Might be because I'm on WP 4.5 beta4. Let me know if it is and I'll look into it.
- if you open a widget by clicking in the sidebar and immediately close it again without making any changes, a refresh is triggered. This doesn't happen if the widget is opened by clicking on it in the preview.
- highlighting of widgets in the sidebar interface doesn't stand out very much. (at not least on my laptop)

js/siteorigin-panels/dialog/widget.js
gregpriday: I've been pretty bad about style guidelines, but I think we should follow the WordPress JS conventions. They're pretty similar to the PHP coding guidelines, so easier to keep in mind.

At some point, I'd like to do a complete run through all our code to fix style. Even running it through an PHPStorm code formatting would help.

js/siteorigin-panels/view/builder.js
gregpriday: Yeah, much neater. Will go with that.

tpl/js-templates.php
gregpriday: Aaah yeah, strictly this should be `esc_html_e`. That's a slightly better escaping to use where there shouldn't be any HTML added in the translation. `_e` would definitely work though, and probably more than secure enough.

gregpriday:
> CPT plugin doesn't load PB interface. Might be because I'm on WP 4.5 beta4. Let me know if it is and I'll look into it.

Yeah, I think with all the changes we've made lately, we probably need to get the CPT plugin back up to speed. Let's focus on that after the 2.4 release of Page Builder out. Luckily we still haven't released the CPT plugin.

> if you open a widget by clicking in the sidebar and immediately close it again without making any changes, a refresh is triggered. This doesn't happen if the widget is opened by clicking on it in the preview.

That's because of the `raw` flag that we have somewhere. As soon as a widget is edited on the frontend, a raw flag is added. This lets the backend know that it's dealing with raw form data, which needs to be passed through the widgets `update` function. You're right though, we should make sure this doesn't trigger a change.

> highlighting of widgets in the sidebar interface doesn't stand out very much. (at not least on my laptop)

Good call.

Good stuff, a few more things to change. I made a few small changes while I was offline too, but going to implement these changes then update the PR.

js/siteorigin-panels/view/live-editor.js
gregpriday: I think this was so we could have the wrapper in the actual template instead. Having a set, named wrapper was important for how we're moving the builder in and out of the live editor sidebar.

gregpriday: Unless I'm mistaken, render should only be called once?

js/siteorigin-panels/dialog/widget.js
braamgenis: Ah right, the WordPress JS conventions are largely derived from the jQuery style guide. Cool, I'll create an issue for this in the PB repo.

js/siteorigin-panels/view/live-editor.js
braamgenis: In theory, it could be called multiple times, every time the view's model changes. So it might not currently be happening here, but say we want to allow the live editor to preview changes without reloading, then we'd probably need to make changes to the model and re-render the view.

Also, something I missed: It's considered good practice to `return this` at the end of a `render` method.

js/siteorigin-panels/view/live-editor.js
gregpriday: How about `$braam'sIframeContainer`? Yeah, you're right. I remember actually forgetting what that was.

js/siteorigin-panels/view/live-editor.js
gregpriday: Aaah yeah, forgot about the `return this;`. This view should only be getting rendered once because it's not associated with a model, but from a good-practices point of view, I'd be happy to change. So far all `render` functions in PB are similar to this, so would be a pretty big change for all of them. Will look into it.

gregpriday:
Did a few of the fixes. Also added a progress bar to the live editor preview. Nice to offer some reload time indication until we're able to implement partial-reloads for edits.

PR #193, opened 2016-03-23
Fix issue where Layout Builder widget wasn't saving all fields.

braamgenis:
To resolve Widgets Bundle [issue 146](https://github.com/siteorigin/so-widgets-bundle/issues/146).

@gregpriday Very small change, but would just like your input on potential impact.


gregpriday:
Don't see any potential problems with this. Haven't tested it, but if it's working for you, then we're good to merge.

PR #255, opened 2016-07-05
fix xss vulnerability

chakzefir:
Vulnerable file **/siteorigin-panels/tpl/admin-home-page.php** do not properly escape **REQUEST_URI** that makes possible XSS attack.
Add ```esc_url()``` filtration to fix current vulnerability.


gregpriday:
Looking good - thanks!

chakzefir:
@gregpriday could include current fix in nearest release ?

PR #275, opened 2016-08-17
Added filter for column count field

stefan-meier:


gregpriday:
Hi @stefan-meier - sorry for the late reply here. What would the use case for this be? It feels a little bit out of place to me, but if it has a good use case, then we could consider it.

stefan-meier:
Hi @gregpriday, I had to modify the maximum number of columns for a client to make sure editors could only use four columns at max (more didn't make sense because of the layout we were using). I thought maybe it could be helpful for someone else as well.

braamgenis:
Closing old PR as stale. Feel free to open a new issue for discussion.

PR #291, opened 2016-09-13
Object cache the rendered content

joehoyle:
Currently whenever the content or excerpt is called for a post, the
siteprigin plugin has to render the full widget content (and also
enqueue all the scripts and styles for the widgets). This is _very_ slow
(somewhere in the 300ms range, depending on plenty of things of course.)
Given the post content will not change until the post it's self is
updated, it makes sense to make use of the Object Cache API.

It looks like there was some in-memory cache, however this never
functioned, as it appears that there was only code to check the cache,
never populate it.

Because of the scripts and styles that can go with widgets, we have to
use a hacky complex solution to capture all the enqueued scripts and
styles while the widgets are being rendered. Unfortunatly this means
hacking with the globals, but I don't know of a better way to acheive
this other than a more declerative API for widget scripts and styles.

The patch was tested and singificantly improved page render time in my
testing (from about 800ms to 400ms) however this is going to depend a
lot on what widgets are used for example.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -939,6 +939,15 @@ function siteorigin_panels_filter_content( $content ) {
 }
 add_filter( 'the_content', 'siteorigin_panels_filter_content' );
 
+/**
+ * Flush the siteorigin content cache when a post is updated.
+ *
+ * @param  int $post_id
+ */
+function siteorigin_panels_flush_post_cache( $post_id ) {
+	wp_cache_delete( $post_id, 'siteorigin_panels' );
+}
+add_filter( 'clean_post_cache', 'siteorigin_panels_flush_post_cache' );
 
 /**
  * Render the panels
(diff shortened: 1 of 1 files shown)


joehoyle:
Sad times 😢

gregpriday:
Hi @joehoyle - apologies for missing this. I'll give it a full review and merge it for the 2.5 release. It looks like an awesome addition.

gregpriday:
So the main challenge with this implementation is that it doesn't allow for dynamic widget content. For example, let's say there's a date widget on the page. This would always display the date at the time the cache is generated. The same would apply for things like Twitter widgets.

This means that cache would need to be optional, or we'd need to limit what is cached.

Also, to get around using globals, I'd prefer this to be implemented using a class structure. We're moving to this structure for [version 3.0 of Page Builder](https://github.com/siteorigin/siteorigin-panels/tree/versions/3.0).

joehoyle:
@gregpriday yes this is true - if you consider the page content to be collection of widgets, more like a sidebar I can see the limitation. In the "page builder" context I think there's a _little_ bit more of an expectation of static content, however even saying that now, I'm not so swayed by this.

This is a bit of a sledgehammer fix to the performance issues in the page builder, I'm sure it's _possible_ to have most of the widgets dynamic, but due with the current implementation, doing so is bad for performance.

I'm not sure if it makes sense for this to be merged, but I thought I'd open the PR with the code I used to fix this performance issue in our case (were dynamic widgets are not an issue).

gregpriday:
So one option would be to only cache when the page only consists of Editor
and Text widgets. It would just need to skip shortcode processing on
caching, and process them after retrieving from cache.

That would cover that fairly common case of static pages.

Thinking about it, even dynamic widgets could be replaced with a special
shortcode in the cache. I imagine shortcode processing is a lot faster than
full Page Builder rendering.

gregpriday:
@joehoyle - I spent some time yesterday going over caching Page Builder content. There are quite a few challenges. The ones I mentioned in my previous message plus a few other potential issues like dynamic styling, the sidebar emulator, etc.

So for the core of Page Builder, I'm going to add a few filters that would facilitate a caching plugin. Essentially filters that allow other plugins to provide the CSS, HTML and even scripts/styles. That way, we can create a separate experimental caching plugin.

gregpriday:
Going to close this PR. I actually tried implementing this feature in https://github.com/siteorigin/siteorigin-panels/commit/dab5ee97ed9eeb3f9f634e66f405dfc11a7a36c8, but in the end we had to [remove it](https://github.com/siteorigin/siteorigin-panels/commit/afbc47b7ba6f307420303417f0abae3eaabe6186). It brought up too many conflicts and issues to make it worth having.

In the future, it'd be great to have this as a separate plugin so that users understand that things might not work properly when caching their content.

PR #295, opened 2016-09-26
Apply row style bottom margin to widgets

### LINKED ISSUE #250: Row style bottom margin doesn't apply to widgets when there are more than one widget in a column
As the title suggests if you set a bottom margin of say, 0 and there are multiple widgets in a column the bottom margin will only apply to the last widget. All other widgets will have the global bottom margin default rather than the desired bottom margin of 0. The no bottom margin setting (in the layout settings group) has no effect either.

Here my [Example layout](https://u.teknik.io/dEOjA.json) which is used in all of the below screenshots.

Screenshot of bottom margin being applied to widget in a multi-widget column.

No bottom margin being applied to the last widget in that column:

No margin being applied in the column to the left when no other widgets are present:

Adjusted global bottom margin (from 35 to 10) setting and change was reflected:

AlexGStapleton:
Resolves #250 

If you set a bottom margin of say, 0, it won't be applied to any of the widgets. This contradicts the global bottom setting as that setting _will_. This PR aims to correct this by applying the row defined bottom margin if set, and if not, defaulting to the global default.

This will not affect the last-child bottom margin.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -3,7 +3,7 @@
 Plugin Name: Page Builder by SiteOrigin
 Plugin URI: https://siteorigin.com/page-builder/
 Description: A drag and drop, responsive page builder that simplifies building your website.
-Version: 2.4.15
+Version: dev
 Author: SiteOrigin
 Author URI: https://siteorigin.com
 License: GPL3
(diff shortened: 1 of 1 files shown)


KevinTaron:
Hey @AlexGStapleton 

i've the same problem. 
Sometimes I don't want the widget to have margin-bottom.

But I would like have an option on each widget so the user can decide if he is removing the margin or not.

Something like this:

default-style.php - line 186:
		
$fields['bottom_margin'] = array(
			'name' => __('Bottom Margin', 'siteorigin-panels'),
			'type' => 'measurement',
			'group' => 'layout',
			'description' => sprintf( __('Space below the row. Default is %spx.', 'siteorigin-panels'), siteorigin_panels_setting( 'margin-bottom' ) ),
			'priority' => 5,
		);

----

But the user would have to calculate the margin-bottom for the cell minus the general margin-bottom.
So if I want to have no margin in my cell i would have to set -30px because -30px + default (30px) = 0

@gregpriday maybe there is a better solution?

AlexGStapleton:
That's kind of a separate issue but I think the best solution would be to have a margin-bottom field for the widget itself. If it's not set, either the row or global setting will be used. If it is, that value will be used.

KevinTaron:
Yeah that would be awesome.
Global Setting < Row Setting < Widget Setting

AlexGStapleton:
Closed in favor of introducing this to 3.x branch.

PR #300, opened 2016-10-20
Replace TinyMCE _.isUndefined()  check with a typeof

AlexGStapleton:
_.isUndefined() returns a reference error if TinyMCE isn't set ([reference](https://github.com/jashkenas/underscore/issues/250)). This means that if the visual editor is disabled (which can be done via the user profile) SiteOrigin Page Builder won't save any changes. Changing _.isUndefined() to a typeof check resolves this issue and allows the user to continue using SiteOrigin Page Builder.

### DIFF
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -689,7 +689,7 @@ module.exports = Backbone.View.extend( {
 	 */
 	updateEditorContent: function ( content ) {
 		// Switch back to the standard editor
-		if ( this.config.editorType !== 'tinymce' || typeof tinymce === 'undefined' || _.isNull( tinyMCE.get( "content" ) ) ) {
+		if ( this.config.editorType !== 'tinyMCE' || typeof tinymce === 'undefined' || _.isNull( tinyMCE.get( "content" ) ) ) {
 			var $editor = $( this.config.editorId );
 			$editor.val( content ).trigger( 'change' ).trigger( 'keyup' );
 		} else {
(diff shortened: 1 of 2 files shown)


js/siteorigin-panels/view/builder.js
```diff
@@ -689,7 +689,7 @@ module.exports = Backbone.View.extend( {
 	 */
 	updateEditorContent: function ( content ) {
 		// Switch back to the standard editor
-		if ( this.config.editorType !== 'tinymce' || _.isUndefined( tinyMCE ) || _.isNull( tinyMCE.get( "content" ) ) ) {
+		if ( this.config.editorType !== 'tinymce' || typeof tinymce === 'undefined' || _.isNull( tinyMCE.get( "content" ) ) ) {
```
gregpriday: Are `tinyMCE` and `tinymce` the same thing? For consistency, we should stick to a single way of accessing it. So `tinyMCE`.

AlexGStapleton: They are, which explains why I didn't pick on this.

I'll switch them over to TinyMCE now.

AlexGStapleton:
Did you see my follow up? Were there any other issues?

js/siteorigin-panels/view/builder.js
```diff
@@ -689,7 +689,7 @@ module.exports = Backbone.View.extend( {
 	 */
 	updateEditorContent: function ( content ) {
 		// Switch back to the standard editor
-		if ( this.config.editorType !== 'tinymce' || _.isUndefined( tinyMCE ) || _.isNull( tinyMCE.get( "content" ) ) ) {
+		if ( this.config.editorType !== 'tinyMCE' || typeof tinymce === 'undefined' || _.isNull( tinyMCE.get( "content" ) ) ) {
```
gregpriday: There's just one more instance of using `tinymce` on this line.

AlexGStapleton:
Oh. I think I need an eye test as I have no idea how I missed that. 😖

PR #309, opened 2016-11-07
Feature/filter inside row

mattrobertsky:
Hi there

Great plugin, thanks for all your hard work on it :)

We have added a filter to allow us to add wrapper elements inside rows. We'd love to get this into your next release. Can you let us know if this is possible and what your timeframe might be like.

many thanks, Matt

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -1085,6 +1085,9 @@ function siteorigin_panels_render( $post_id = false, $enqueue_css = true, $panel
 		$row_style_wrapper = siteorigin_panels_start_style_wrapper( 'row', $style_attributes, !empty($panels_data['grids'][$gi]['style']) ? $panels_data['grids'][$gi]['style'] : array() );
 		if( !empty($row_style_wrapper) ) echo $row_style_wrapper;
 
+		// Themes can add a wrapper within the row itself
+		echo apply_filters('siteorigin_panels_inside_row_start', '');
+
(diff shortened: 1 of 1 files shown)


gregpriday:
Thanks for the suggestion @mattroberts-brandwatch! So the main thing with a filter like this is that we'd need to make sure the live editor can properly handle any additional HTML added in these filters. It might work straight away - I haven't actually checked yet.

One last thing we need to consider is the naming of filters. Ideally, we'd look at all the places that we could add filters, and add them all at once with a consistent naming convention. There's an existing filter called `siteorigin_panels_before_row`, which feels different to `siteorigin_panels_inside_row_start` - a better name might be `siteorigin_panels_before_row_content` and `siteorigin_panels_after_row_content`.

Right now, I'm focusing most of my time on a major 3.0 update of Page Builder. I'd actually prefer to see any new filters added to this release.

https://github.com/siteorigin/siteorigin-panels/tree/versions/3.0

mattroberts-brandwatch:
Hi @gregpriday, thanks for getting back to me. Am very happy to change the name, I'll make a PR on the release branch if that is better? Can I ask roughly when you think 3.0 will be released?

gregpriday:
Going to close this PR for now. Sorry we didn't get to it sooner. At the moment there are merge conflicts and we haven't had a need for this particular filter. If you can solve the conflicts and resubmit, we'll take a look. If you no longer need this, that's fine too.

PR #320, opened 2016-12-13
PB 3.0: Add ability to filter widget title

AlexGStapleton:
This will allow users to filter the widget title (if present). It'll also allow users more context when using siteorigin_panels_widget_args as to what the widget title actually is.


gregpriday:
@AlexGStapleton - can you resubmit this to the develop branch, for Page Builder 2.5.

PR #323, opened 2016-12-19
PB 3.0: Replace TinyMCE _.isUndefined() check with a typeof

AlexGStapleton:
Copy of #300.


gregpriday:
I've cherry picked all changes we need from the 3.0 branch into develop. We're going to make these changes in a few separate stages, rather than all at once.

PR #325, opened 2016-12-28
PB 3.0: is_live_editor()

AlexGStapleton:
This PR adds is_live_editor() and a body_class.

body_class - Allow designers to hide certain elements that can cause issues when viewed via the live editor

conditional check: Allows for developers to easily output live editor specific JavaScript and styling

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -136,11 +136,7 @@ class SiteOrigin_Panels {
 	 * @return bool
 	 */
 	public static function  is_live_editor(){
-		if ( ! empty( $_GET['siteorigin_panels_live_editor'] ) ) {
-			return true;
-		} else {
-			return false;
-		}
+		return ! empty( $_GET['siteorigin_panels_live_editor'] )
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


siteorigin-panels.php
```diff
@@ -129,6 +129,19 @@ public static function is_panel( $can_edit = false ) {
 		$is_panel =  ( siteorigin_panels_is_home() || ( is_singular() && get_post_meta(get_the_ID(), 'panels_data', false) ) );
 		return $is_panel && ( ! $can_edit || ( ( is_singular() && current_user_can( 'edit_post', get_the_ID() ) ) || ( siteorigin_panels_is_home() && current_user_can('edit_theme_options') ) ) );
 	}
+	
+	/**
+	 * Check if the user is currently using the live editor.
+	 *
+	 * @return bool
+	 */
+	public static function  is_live_editor(){
```
gregpriday: This could be done a lot more efficiently with just

`return ! empty( $_GET['siteorigin_panels_live_editor'] )`

AlexGStapleton: Oh... you're right. Sorry about that.

Updated.

Issue #82, opened 2015-07-28
"Section" container addition


brentjett:
Hi, nice solid plugin. I'd love to see the addition of a "section" container element to optionally wrap around rows/columns. It's very common for multiple rows to share the same background and therefore need a containing element, likely which will span the full width of the page. For example:

![section-example](https://cloud.githubusercontent.com/assets/1790660/8945259/83ab9c04-354b-11e5-9e94-3a4a342227b9.jpg)

Finetuned:
+!

ste93cry:
:+1: I'm using this plugin for the first time and I'm missing this feature a lot!

gregpriday:
This would be fairly easy to implement now using a Layout Widget inside a row with a standard Full Width layout.

ste93cry:
It should be possible to specify the CSS classes for the container `div` in the backend, or at least with a filter

brentjett:
It's no problem, I use Beaver Builder now 👍

gregpriday:
@ste93cry Aaah, @AlexGStapleton let me know I misunderstood the request. It's essentially CSS column styling for text. This is the kind of thing we could add to the Editor Widget in the Widgets Bundle. I'll take a look at this.

@brentjett awesome, thanks for stopping by to let us know.

PR #328, opened 2017-01-11
Add option to add a container to rows

rideron89:
This adds an option in the Row Layout settings to 'Contained'. When this is selected, a container is placed inside the row so a max-width can be specified in the global Page Builder options. This is for rows that don't need content to stretch to full-width.

The default max-width is set to 860px, but can be configured in **Settings** > **Page Builder** > **Layout** > **Row Container Width**.

This may be a fix for the old #82 ticket.

### DIFF
--- a/css/front.css
+++ b/css/front.css
@@ -47,3 +47,17 @@
   display: table;
   clear: both;
 }
+.panel-row-container {
+  margin: 0 auto;
+  width: 100%;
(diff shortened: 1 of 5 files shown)


gregpriday:
Gave some feedback here https://github.com/siteorigin/siteorigin-panels/issues/82#issuecomment-287697610

PR #337, opened 2017-02-03
The "Insert" options are not visible

Smartik89:
On "Import/Export" section the "Insert" button has 3 options that are not visible because the z-index `.so-toolbar` is equal with `so-content`. See the screenshot, it says all.

screen-0003


gregpriday:
Strangely, I'm not seeing this issue. It definitely looks like it should be happening, though.

PR #344, opened 2017-03-01
Cell styles WIP

braamgenis:
Still needs some spit 'n polish, but creating the PR so long for easier initial review and feedback.

### DIFF
--- a/inc/styles.php
+++ b/inc/styles.php
@@ -37,7 +37,8 @@ class SiteOrigin_Panels_Styles {
 				break;
 
 			case 'cell':
-				$this->render_styles_fields( 'cell', '<h3>' . __( 'Cell Styles', 'siteorigin-panels' ) . '</h3>', '', $current, $post_id, $args );
+				$cell_number = isset( $args['index'] ) ? ' ' . ( intval( $args['index'] ) + 1 ) : '';
+				$this->render_styles_fields( 'cell', '<h3>' . sprintf( __( 'Cell%s Styles', 'siteorigin-panels' ), $cell_number ) . '</h3>', '', $current, $post_id, $args );
                 break;
 
 			case 'widget':
--- a/inc/styles.php
+++ b/inc/styles.php
@@ -11,6 +11,7 @@ class SiteOrigin_Panels_Styles {
 
 	public static function single() {
 		static $single;
+
 		return empty( $single ) ? $single = new self() : $single;
 	}
 
@@ -39,7 +40,7 @@ class SiteOrigin_Panels_Styles {
 			case 'cell':
 				$cell_number = isset( $args['index'] ) ? ' ' . ( intval( $args['index'] ) + 1 ) : '';
 				$this->render_styles_fields( 'cell', '<h3>' . sprintf( __( 'Cell%s Styles', 'siteorigin-panels' ), $cell_number ) . '</h3>', '', $current, $post_id, $args );
-                break;
+				break;
 
 			case 'widget':
 				$this->render_styles_fields( 'widget', '<h3>' . __( 'Widget Styles', 'siteorigin-panels' ) . '</h3>', '', $current, $post_id, $args );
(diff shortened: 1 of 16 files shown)


gregpriday:
Awesome, looking good so far. Here's current feedback:

* We need some more indication of selected cell in the edit row dialog. So the heading for cell styles should say "Cell 1 Styling", etc. Also, to match the main interface, it should highlight the currently selected cell. You can check the main interface to see the colours to use for highlighting (the darker blue).
* Is there any way to store the cell styling form? So they don't need to be reloaded each time. For example, when you open a row edit dialog, the row styling form only loads the first time.
* It looks like the add row in the contextual menu doesn't work. Should be a fairly simple fix. Just make sure any changes you made to the add row function are backwards compatible with the old way.

That's all I can see for now, but I'll carry on playing around and do a proper code review.

inc/styles.php
```diff
@@ -36,8 +36,14 @@ function action_style_form() {
 				$this->render_styles_fields( 'row', '<h3>' . __( 'Row Styles', 'siteorigin-panels' ) . '</h3>', '', $current, $post_id, $args );
 				break;
 
+			case 'cell':
+				$cell_number = isset( $args['index'] ) ? ' ' . ( intval( $args['index'] ) + 1 ) : '';
+				$this->render_styles_fields( 'cell', '<h3>' . sprintf( __( 'Cell%s Styles', 'siteorigin-panels' ), $cell_number ) . '</h3>', '', $current, $post_id, $args );
+                break;
```
gregpriday: They call me Greg, the space indentation finder.

js/siteorigin-panels/dialog/row.js
```diff
@@ -174,8 +175,10 @@ module.exports = panels.view.dialog.extend( {
 				handle
 					.appendTo( newCell )
 					.dblclick( function () {
-						var t = thisDialog.row.cells[i] + thisDialog.row.cells[i - 1];
-						thisDialog.row.cells[i] = thisDialog.row.cells[i - 1] = t / 2;
+                        var prevCellModel = thisDialog.row.cells.at(i - 1);
```
gregpriday: More space indents messing with our code style. I never see these in the editor, but on Github they make the code look strange. There are a few others in this file. Could just do a Code Reformat to fix all of them.

gregpriday:
Something I just found in testing. The cell styling settings don't save if you close, then reopen the dialog.

* Open a row settings dialog.
* Select a cell and change its background color.
* Close the dialog.
* Reopen the same row settings, and check the cell's background color. It'll be blank.

gregpriday:
Manually setting the row weights seems to be conflicting with the click event for selecting a cell.

* Open a row settings dialog.
* Click on one of the weight values displayed inside a cell.
* Change the value.
* Press enter or click outside the cell. Both of these events should cause the cells to reweight. This doesn't seem to be working, though.
* Also, with a value selected, pressing tab should take you to the next cell.

You can check `develop` for an idea on how this should work.

braamgenis:
Cool cool. Will get these fixed up.

gregpriday:
Getting there. There are a couple more issues. The first is minor, the headings for cell styling no longer indicate which cell is selected. They just say "Cell Styles".

It's also possible to have to (maybe more) cell style blocks loaded at a time. Steps to recreate: 
* Edit an existing row.
* Click on a cell.
* Close the edit row dialog, then reopen it 
* Click on a cell, you'll now have a repeated cell styles section.

I also can't tell if [this issue](https://github.com/siteorigin/siteorigin-panels/pull/344#issuecomment-284637821) is fixed, because the new cell style area doesn't have the previously entered cell styles.

A usability issue. Clicking on a cell should always select it.
* Open an edit row dialog.
* Click on a cell to select it.
* Click on any of the row widths.
* Click on any cell.

Clicking on a cell in the last step causes it be deselected. Clicking on another cell (outside the cell width field), should lock in the row width values, and ensure that the cell the user clicked is active.

I think that's it. Just do a quick check over all the new functionality to make sure that no fixes break previous fixes.

braamgenis:
I can't reproduce the repeated styles view issue. Could you give more detail on how to reproduce?

PR #351, opened 2017-03-14
Small refactor of render function

braamgenis:
This simply reduces the size of the render function and makes things a bit easier to follow by separating some of the processing code and rendering of rows, cells, and widgets into their own functions.


braamgenis:
Hrm... not sure why there's a conflict. I merged develop into this branch locally without problems.

PR #352, opened 2017-03-14
Small refactor of render function

braamgenis:
This simply reduces the size of the render function and makes things a bit easier to follow by separating some of the processing code and rendering of rows, cells, and widgets into their own functions.

(Resubmission of #351 due to conflict that I wasn't able to resolve. I think it was due to 07c05957dca4e0f35665ccd7da384e84610fbce8 being present in both the feature branch and develop)

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -193,6 +193,7 @@ class SiteOrigin_Panels_Renderer {
 	 * @return string
 	 */
 	function render( $post_id = false, $enqueue_css = true, $panels_data = false ) {
+
 		if ( empty( $post_id ) ) {
 			$post_id = get_the_ID();
 		}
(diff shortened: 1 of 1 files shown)


inc/renderer.php
```diff
@@ -622,4 +458,198 @@ function enqueue_styles() {
...
+			if ( post_password_required( $post_id ) ) {
+				return false;
+			}
+			$panels_data = get_post_meta( $post_id, 'panels_data', true );
+		}
+
+		return $panels_data;
+	}
+
+	private function migrate_widgets_data( $panels_widgets ) {
```
gregpriday: At some point, we should actually look at removing this migration. I think we changed to panels_info before releasing Page Builder 2.0. This doesn't need to happen now, but we should earmark it for removal.

It is performing one step we still need on [line 502](https://github.com/siteorigin/siteorigin-panels/pull/352/files#diff-c80be39f16daedc9fdb681582ecb3432R502).

inc/renderer.php
```diff
@@ -622,4 +458,198 @@ function enqueue_styles() {
...
+				}
+
+				$panels_widgets[ $i ]['panels_info']['widget_index'] = $i;
+			}
+		}
+
+		return $panels_widgets;
+	}
+
+	private function get_panels_layout_data( $panels_data ) {
```
gregpriday: A phpdoc comment for this function would be useful. You know, so we don't lose it again.

inc/renderer.php
braamgenis: Ah, OK cool. I can move that line to `get_panels_layout_data` and remove.

braamgenis:
@gregpriday Cool, added the PHPDocs and removed the `migrate_widgets_data` function.

gregpriday:
Comments make me happy.

PR #354, opened 2017-03-15
Copy paste feature

### LINKED ISSUE #5: Feature request - ROW Copy & Paste
Hi Greg,

I love Page Builder, what it does and what it can do!

This enhancement can save people a lot of time and make it easier to use & re-use developed work.

**ROW Copy & Paste**
- Function: **COPY entire row as is** (including all formatting & widgets)
- Function: **PASTE entire row copied** _to another Page/Site_
- Ideal would be able to _Copy & Paste from one site to another_
- But within a site would be helpful also

Benefits
- This will allow developers to have a **Page with Model Rows** configured to there developed liking
- Then they can just Copy & Paste rows into the Site/Pages they are developing!
- Big time saver & easily allows re-use.

Note ... It is up to the developer to ensure (relatively minor stuff for us) that
-  Widgets are available on site they are copying to
-  URLs are updated as needed
-  etc. 

The Way it is now ... It is All or Nothing!
- To add a row into a new page either you Create the row & widget configs from scratch
  
  -or-
- Clone a whole page in (from current site).

If you can do it, Please consider this.  I believe it will be a big boost & will make Page Builder even easier to use and more productive!

People will be able to **Develop a Row once**, **cherry pick Rows** as needed, then **tweak Rows** as necessary.  
And because developing sites will be even easier, Page Builder will more likely be chosen!

Thanks Greg ... Great job!
Justin

gregpriday:
This resolves #5 by using `localStorage`, rather than the user's clipboard. When there's a more robust, cross browser solution to accessing the user's clipboard, then we'll switch to using that.


js/siteorigin-panels/model/builder.js
```diff
@@ -17,7 +17,7 @@ module.exports = Backbone.Model.extend({
 
     initialize: function () {
         // These are the main rows in the interface
-        this.rows = new panels.collection.rows();
+		this.set( 'rows', new panels.collection.rows() );
```
braamgenis: Seems this whole file needs to be reformatted to use tabs instead of spaces.

js/siteorigin-panels/view/builder.js
```diff
@@ -500,28 +503,18 @@ module.exports = Backbone.View.extend( {
...
-			if ( options.defaultPosition === 'last' ) {
-				activeCell = this.$( '.so-cells .cell' ).first();
-			} else {
-				activeCell = this.$( '.so-cells .cell' ).last();
-			}
 		}
 
-		return activeCell.data( 'view' ).model;
+		var activeCell = this.activeCell;
+		return _.isEmpty( activeCell ) || this.model.get('rows').indexOf(activeCell.model) === -1 ? this.model.get('rows').last().get('cells').first() : activeCell.model;
```
braamgenis: Loooong line is long. :stuck_out_tongue: It's a bit hard to see that the expression

`_.isEmpty( activeCell ) || this.model.get('rows').indexOf(activeCell.model) === -1` 

is what's being checked for the ternary. So perhaps to reduce length of the line and make it more readable extract that expression into a variable before the `return`.

js/siteorigin-panels/view/builder.js
```diff
@@ -500,28 +503,18 @@ module.exports = Backbone.View.extend( {
...
 			} else {
 				return null;
 			}
-
-		}
-
-		var activeCell = this.$( '.so-cells .cell.cell-selected' );
-
-		if ( activeCell.length === 0 ) {
-			if ( options.defaultPosition === 'last' ) {
```
braamgenis: If we're not taking `defaultPosition` option into account anymore then it should probably be removed from line 503 too?

js/siteorigin-panels/view/row.js
```diff
@@ -161,14 +161,47 @@ module.exports = Backbone.View.extend( {
...
 	/**
+	 * Copy the row to a localStorage
+	 */
+	copyHandler: function(){
+		if ( typeof(Storage) === "undefined" || ! panelsOptions.user ) return;
+
+		var serial = panels.serial.serialize( this.model );
+		serial.thingType = 'row-model';
+
+		// Store this in the cookie
```
braamgenis: What cookie? :cookie:

js/siteorigin-panels/view/row.js
```diff
@@ -161,14 +161,47 @@ module.exports = Backbone.View.extend( {
...
+		localStorage[ 'panels_clipboard_' + panelsOptions.user ] = JSON.stringify( serial );
+	},
+
+	/**
+	 * Create a new row and insert it
+	 */
+	pasteHandler: function(){
+		if ( typeof(Storage) === "undefined" || ! panelsOptions.user ) return;
```
braamgenis: In general I feel like the specifics of the copy-paste implementation doesn't belong in the view, but should rather be separated into a model or helper. The copy and paste handler functions should just be saying something like `var pastedModel = clipboardHelper.getPastedModel();` and not care how that pastedModel is retrieved or rebuilt. No need to change now, but I think we should consider a small refactor for that later.

js/siteorigin-panels/view/row.js
gregpriday: Old comments :(

js/siteorigin-panels/view/row.js
gregpriday: Yeah, agree there. It would also make it much easier to change the backend of the clipboard (for example, to an actual clipboard).

gregpriday: @braamgenis hopefully this is what you meant - https://github.com/siteorigin/siteorigin-panels/pull/354/commits/70c779c1e73d6853705ddf223d399cd95309f405

braamgenis: Yep, that's what I had in mind. Looks good. :+1:

PR #364, opened 2017-03-21
Editor customizations WIP

braamgenis:
Row colors still need to be applied to border and resize handle.
Going to add the widget titles editing here too.


gregpriday:
First I'll give general feedback on the row color labelling. Because this is a style based thing, I'd much rather see all color labels handled with CSS classes. This means the row will only store a single color name like `red`, `blue`, etc. This will translate to a class like `so-color-label-red` on the main row wrapper, which will in turn style the row to give it the color label.

This does mean we can't give users the option to add custom color labels, but it feels a lot more robust. It also takes care of visual styling widgets as they're dragged in and out of cells. Which is something the design would call for.

For bonus points, try have the same CSS style both the buttons in the dropdown menu, and the row itself. This will make it much easier to add/modify color labels.

So we're basing this update on suggestions from Proteus Themes. Here's the image they sent us.

page_builder_suggestions

For now, let's just go with the colors they suggested in the design. I'll look at adding more before we release.

* Notice that we're dropping the border around the cells.
* The border around widgets is based on the row color label. You can see this by zooming into the image.
* Each row has a default blue color label selected, which is the original light blue color. This gives users a better idea of what the color label actually does. 
* The colors they've suggested are
   * Original:`#e4eff4` and `#9bafb5`
   * Purple:`#ebe4f4` and `#a49bb5`
   * Yellow:`#f7f0e0` and `#bbac95`
   * Turquoise:`#ddf9f3` and `#90bfb0`
* Notice the styling of the check mark. It has a more square appearance and some padding around it.

gregpriday:
Feedback on the widget labelling - this needs a little more indication of what's happening. So when the user hovers over the label, it should highlight the background color to a subtly darker grey to imply that clicking there will have some action on the label.

We should also remove a lot of the styling on the text input to make sure there isn't a jump when the user clicks on the input. This would also mean we don't have to remove the sub-headline for a widget when editing the main headline.

I'll do a proper code review once all the changes are in.

js/siteorigin-panels/view/dialog.js
```diff
@@ -345,6 +351,34 @@ module.exports = Backbone.View.extend( {
...
 	/**
+	 * Start editing the title.
+	 */
+	editTitle: function( event ) {
+		if ( this.editableTitle ) {
+			var titleElt = this.$( '.so-title' );
+			var titleInput = this.$( '.so-edit-title' );
+			titleElt.hide();
+			titleInput.show();
+			titleInput.focus();
```
gregpriday: Ideally, this should select all text too.

js/siteorigin-panels/model/widget.js
```diff
@@ -45,6 +45,8 @@ module.exports = Backbone.Model.extend( {
 			} else {
 				return '';
 			}
+		} else if ( this.has( 'title' ) ) {
```
gregpriday: I actually see that it's been called title quite a lot, so I'll let you make this call. I'm actually fine with either title or label. I'd just like to avoid confusion with the title field added by a lot of widgets that's displayed on the frontend.

PR #365, opened 2017-03-21
Move style CSS into general CSS object

gregpriday:
This is a fairly large refactoring of how row/cell/widget styling works. I've renamed a few things and added a few filters to make it easier for developers to add their own styling CSS.

### DIFF
--- a/inc/default-styles.php
+++ b/inc/default-styles.php
@@ -7,54 +7,40 @@
  */
 class SiteOrigin_Panels_Default_Styles {
 
-	static function init() {
-		add_action( 'wp_enqueue_scripts', array( 'SiteOrigin_Panels_Default_Styles', 'register_scripts' ), 5 );
+	public function __construct() {
+		add_action( 'wp_enqueue_scripts', array( $this, 'register_scripts' ), 5 );
 
 		// Adding all the fields
-		add_filter( 'siteorigin_panels_row_style_fields', array(
-			'SiteOrigin_Panels_Default_Styles',
-			'row_style_fields'
-		) );
+		add_filter( 'siteorigin_panels_row_style_fields', array( $this, 'row_style_fields' ) );
+		add_filter( 'siteorigin_panels_cell_style_fields', array( $this, 'cell_style_fields' ) );
+		add_filter( 'siteorigin_panels_widget_style_fields', array( $this, 'widget_style_fields' ) );
 
-		add_filter( 'siteorigin_panels_cell_style_fields', array(
-			'SiteOrigin_Panels_Default_Styles',
-			'cell_style_fields'
-		) );
+		// Style wrapper attributes
+		add_filter( 'siteorigin_panels_row_style_attributes', array( $this, 'general_style_attributes' ), 10, 2 );
+		add_filter( 'siteorigin_panels_row_style_attributes', array( $this, 'row_style_attributes' ), 10, 2 );
+		add_filter( 'siteorigin_panels_cell_style_attributes', array( $this, 'general_style_attributes' ), 10, 2 );
+		add_filter( 'siteorigin_panels_widget_style_attributes', array( $this, 'general_style_attributes' ), 10, 2 );
 
-		add_filter( 'siteorigin_panels_widget_style_fields', array(
-			'SiteOrigin_Panels_Default_Styles',
-			'widget_style_fields'
-		) );
+		// Style wrapper CSS
+		add_filter( 'siteorigin_panels_row_style_css', array( $this, 'general_style_css' ), 10, 2 );
+		add_filter( 'siteorigin_panels_cell_style_css', array( $this, 'general_style_css' ), 10, 2 );
+		add_filter( 'siteorigin_panels_widget_style_css', array( $this, 'general_style_css' ), 10, 2 );
 
-		// Filter the row style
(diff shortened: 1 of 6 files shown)


inc/renderer.php
```diff
@@ -261,47 +261,49 @@ function render( $post_id = false, $enqueue_css = true, $panels_data = false ) {
...
 			// Filter and sanitize the classes
-			$style_attributes['class'] = apply_filters( 'siteorigin_panels_' . $name . '_style_classes', $style_attributes['class'], $style_attributes, $style_args );
-			$style_attributes['class'] = array_map( 'sanitize_html_class', $style_attributes['class'] );
+			$attributes['class'] = apply_filters( 'siteorigin_panels_' . $name . '_style_classes', $attributes['class'], $attributes, $style );
+			$attributes['class'] = array_map( 'sanitize_html_class', $attributes['class'] );
 
 			$style_wrapper = '<div ';
-			foreach ( $style_attributes as $name => $value ) {
+			foreach ( $attributes as $name => $value ) {
+				if( substr( $name, 0, 1 ) === '_' ) continue;
```
braamgenis: Is there a special type of attribute that starts with an underscore? Something I'm usually quite bad with is leaving clarifying comments  for special cases like this, so will try to point these out when I see them and fix them in my own code too.

inc/default-styles.php
```diff
@@ -65,65 +65,50 @@ static function register_scripts() {
...
+
+		// All the attribute fields
 
 		$fields['id'] = array(
-			'name'        => __( 'Row ID', 'siteorigin-panels' ),
+			'name'        => sprintf( __( '%s ID', 'siteorigin-panels' ), $label ),
 			'type'        => 'text',
 			'group'       => 'attributes',
-			'description' => __( 'A custom ID used for this row.', 'siteorigin-panels' ),
+			'description' => sprintf( __( 'A custom ID used for this %s.', 'siteorigin-panels' ), strtolower( $label ) ),
```
braamgenis: Couldn't this just use `$id` instead of `strtolower( $label )`?

inc/default-styles.php
```diff
@@ -65,65 +65,50 @@ static function register_scripts() {
...
-			'priority'    => 6,
-		);
+		// The layout fields
 
 		$fields['padding'] = array(
 			'name'        => __( 'Padding', 'siteorigin-panels' ),
 			'type'        => 'measurement',
 			'group'       => 'layout',
-			'description' => __( 'Padding around the entire row.', 'siteorigin-panels' ),
+			'description' => sprintf( __( 'Padding around the entire %s.', 'siteorigin-panels' ), strtolower( $label ) ),
```
braamgenis: Same here about using `$id` and all the lines below.

inc/default-styles.php
```diff
@@ -220,102 +160,109 @@ static function row_style_fields( $fields ) {
...
 		return $fields;
 	}
 
-	static function cell_style_fields( $fields ) {
-		$fields['class'] = array(
+	static function row_style_fields( $fields ) {
+		// Add the general fields
+		$fields = wp_parse_args( $fields, self::get_general_style_fields( 'row', __( 'Row', 'siteorigin-panels' ) ) );
+
+		//Should we remove this? Or move it's existing value to new Cell Styles?
```
braamgenis: Can remove this comment. We ended up keeping this as applying to all cells, but it will be overridden by the specific cell styles `class` field.

inc/default-styles.php
```diff
@@ -433,97 +310,119 @@ static function widget_style_fields( $fields ) {
...
-					$attributes['style'] .= $matches[1][ $i ] . ':' . $matches[2][ $i ] . ';';
-				}
+			if ( $style['background_display'] == 'parallax' || $style['background_display'] == 'parallax-original' ) {
+				wp_enqueue_script( 'siteorigin-panels-front-styles' );
 			}
-		}
 
-		if ( ! empty( $args['background'] ) ) {
-			$attributes['style'] .= 'background-color:' . $args['background'] . ';';
+			$url = wp_get_attachment_image_src( $style['background_image_attachment'], 'full' );
```
braamgenis: Unrelated to the purpose of this branch, but we should consider converting all of these to `wp_get_attachment_image_srcset` if we can.

inc/default-styles.php
gregpriday: Technically in all the cases we're using it for, yeah. Although the main difference between the label and the ID is spaces. So say we add some different type of _thing_ to Page Builder that requires a 2 word label. The ID would be `some-thing` and the label would be `Some thing`.

inc/renderer.php
gregpriday: I think this was for a user case that I didn't actually go with. Basically letting another plugin add an attribute starting with `_` that triggers the renderer to add a style div, without adding it to the actual HTML. I don't think it's being used anywhere though.

inc/default-styles.php
gregpriday: Oh rad, and it turns out background images can have a srcset equivalent http://stackoverflow.com/questions/26801745/is-there-a-srcset-equivalent-for-css-background-image - I'll log this as a separate issue, though.

gregpriday: https://github.com/siteorigin/siteorigin-panels/issues/366

PR #369, opened 2017-03-23
Improved interaction with post_content

gregpriday:
This PR improves how Page Builder interacts with post_content. It's now possible to view a Page Builder page, even when the plugin is deactivated. This mainly works with Widgets Bundle widgets.

We can also parse post_content to extract the original Page Builder data.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -857,8 +857,16 @@ class SiteOrigin_Panels_Admin {
 		);
 		$panels_data            = SiteOrigin_Panels_Styles_Admin::single()->sanitize_all( $panels_data );
 
-		define( 'SITEORIGIN_PANELS_DATABASE_RENDER', true );
-		echo SiteOrigin_Panels_Renderer::single()->render( intval( $_POST['post_id'] ), false, $panels_data );
+		$GLOBALS[ 'SITEORIGIN_PANELS_DATABASE_RENDER' ] = true;
+		$post_content = SiteOrigin_Panels_Renderer::single()->render( intval( $_POST['post_id'] ), false, $panels_data );
+		$post_css = file_get_contents( plugin_dir_path( __FILE__ ) . '../css/front.css' );
+		$post_css .= SiteOrigin_Panels_Renderer::single()->generate_css( intval( $_POST['post_id'] ), false, $panels_data );
+		$post_css = preg_replace( '/\s+/', ' ', $post_css );
+
+		echo '<style type="text/css">' . $post_css . '</style>';
+		echo $post_content;
+
+		unset( $GLOBALS[ 'SITEORIGIN_PANELS_DATABASE_RENDER' ] );
 
 		wp_die();
 	}
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -881,7 +881,7 @@ class SiteOrigin_Panels_Admin {
 
 		$GLOBALS[ 'SITEORIGIN_PANELS_DATABASE_RENDER' ] = true;
 
-		do_action( 'siteorigin_panels_setup_database_render' );
+		do_action( 'siteorigin_panels_setup_database_render', $_POST['post_id'] );
 		$post_content = SiteOrigin_Panels_Renderer::single()->render( intval( $_POST['post_id'] ), false, $panels_data );
 		echo SiteOrigin_Panels_Renderer::single()->render( intval( $_POST['post_id'] ), false, $panels_data );
 
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -179,7 +179,9 @@ class SiteOrigin_Panels_Admin {
 
 				// This gives other plugins the chance to
 				do_action( 'siteorigin_panels_setup_database_render', $post_id );
-				SiteOrigin_Panels_Post_Content::single();
+				$filters = SiteOrigin_Panels_Post_Content_Filters::single();
+				$filters->clear_filters();
+				$filters->setup_filters();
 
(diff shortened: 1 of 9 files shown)


inc/admin.php
```diff
@@ -173,10 +172,33 @@ function save_post( $post_id, $post ) {
...
 		if ( ! empty( $panels_data['widgets'] ) || ! empty( $panels_data['grids'] ) ) {
 			update_post_meta( $post_id, 'panels_data', map_deep( $panels_data, array( 'SiteOrigin_Panels_Admin', 'double_slash_string' ) ) );
+
+			if( siteorigin_panels_setting( 'copy-content' ) ) {
+				// Save a copy of all this content into the posts's post_content
+				$GLOBALS[ 'SITEORIGIN_PANELS_DATABASE_RENDER' ] = true;
+
+				// This gives other plugins the chance to
+				do_action( 'siteorigin_panels_setup_database_render', $post_id );
+				SiteOrigin_Panels_Post_Content::single();
```
braamgenis: It's unclear what this does without going to look at the implementation of `SiteOrigin_Panels_Post_Content`. I'd consider renaming it to something like `SiteOrigin_Panels_Post_Content_Filters` or the bigger mouthful `SiteOrigin_Panels_Post_Content_Data_Adapter`. :stuck_out_tongue: And perhaps go a step further and make the call to add the filters an explicit method, instead of tucked away in the constructor.

inc/admin.php
```diff
@@ -857,9 +880,15 @@ function action_builder_content() {
 		);
 		$panels_data            = SiteOrigin_Panels_Styles_Admin::single()->sanitize_all( $panels_data );
 
-		define( 'SITEORIGIN_PANELS_DATABASE_RENDER', true );
+		$GLOBALS[ 'SITEORIGIN_PANELS_DATABASE_RENDER' ] = true;
+
+		do_action( 'siteorigin_panels_setup_database_render', $_POST['post_id'] );
+		SiteOrigin_Panels_Post_Content::single();
+		$post_content = SiteOrigin_Panels_Renderer::single()->render( intval( $_POST['post_id'] ), false, $panels_data );
```
braamgenis: This is rendering content but never uses `$post_content`. I assume you wanted to check the output while debugging and forgot to remove this?

inc/post-content.php
```diff
@@ -0,0 +1,110 @@
...
+
+	public static function single() {
+		static $single;
+		return empty( $single ) ? $single = new self() : $single;
+	}
+
+	/**
+	 * Clear filters so we get predictable HTML
+	 */
+	public function clear_filters(){
```
braamgenis: Looks like this isn't being used anymore, so perhaps it can be removed?

inc/renderer.php
```diff
@@ -34,16 +34,19 @@ public function add_inline_css( $post_id, $css ) {
...
 	 * @return string
 	 */
-	private function generate_css( $post_id, $layout_data, $panels_data ) {
+	public function generate_css( $post_id, $panels_data, $layout_data) {
 		// Exit if we don't have panels data
-		if ( empty( $layout_data ) ) {
+		if ( empty( $layout_data ) && empty( $panels_data ) ) {
 			return;
 		}
+		if( empty( $layout_data ) ) {
```
braamgenis: Just a personal preference thing, but this could probably be combined with the above conditional. Something like 
```
if ( empty( $layout_data ) ) {
    if ( empty( $panels_data ) ) {
        return;
    }
    $layout_data = $this->get_panels_layout_data( $panels_data );
}

File: js/siteorigin-panels/model/builder.js [hunk 5]
```diff
@@ -292,4 +292,174 @@ module.exports = Backbone.Model.extend({
...
+		if( $html.find('.panel-layout .panel-grid').length ) {
+			// This looks like Page Builder html, lets try parse it
+			var panels_data = {
+				grids: [],
+				grid_cells: [],
+				widgets: [],
+			};
+
+			// The Regex object that'll match SiteOrigin widgets
+			// console.log( panelsOptions.siteoriginWidgetRegex );
```
braamgenis: Remove commented out `console.log`

js/siteorigin-panels/model/builder.js
```diff
@@ -292,4 +292,174 @@ module.exports = Backbone.Model.extend({
...
+			var re = new RegExp( panelsOptions.siteoriginWidgetRegex , "i" );
+
+			$html.find('> .panel-layout > .panel-grid').each( function( ri, el ){
```
braamgenis: Would be nice if this could be handled by separate functions, e.g. `parseRow`, `parseCell`, `parseWidget` etc to try reduce the size of this function and make it more readable. To keep the functions bound to the builder model, you can use underscore's [bindAll](http://underscorejs.org/#bindAll) e.g. `_.bindAll(this, 'parseRow', 'parseCell', 'parseWidget');` in `initialize()`.

inc/admin.php
gregpriday: Radical, made those changes. That's a lot clearer.

inc/admin.php
gregpriday: Debugging code ruins lives. Removed this.

inc/post-content.php
gregpriday: To be safe, we'll actually make use of this.

js/siteorigin-panels/model/builder.js
gregpriday: Yeah, that would be a really nice refactoring. I won't do that now, but I'll try to get that done before we release 2.5.

PR #370, opened 2017-03-24
PB 2.5: Add ability to filter widget title

AlexGStapleton:
#320 for 2.5.

Awkwardly, the initial PR applied the title to $args['title'] but it was never actually applied as the title was always taken from $instance['title'] (at least in 99% of cases). As a result, I applied set $instance['title'] to $args['title'] after the `siteorigin_panels_widget_args` hook to allow for plugins/themes and developers to more easily filter it (if desired).

`siteorigin_panels_widget_args` still includes less context, however. In an Ideal world, it would pass $instance.

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -393,8 +393,17 @@ class SiteOrigin_Panels_Renderer {
 			'widget_id'     => 'widget-' . $grid_index . '-' . $cell_index . '-' . $widget_index
 		);
 
+		// Allow plugins/themes to filter widget title
+		if ( !empty( $instance['title'] ) && !empty( $the_widget ) && !empty( $the_widget->id_base ) ) {
+			$args['title'] = apply_filters( 'widget_title', $instance['title'], $instance, $the_widget->id_base );
+		}
+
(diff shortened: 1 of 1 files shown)


gregpriday:
So actually, taking a look at this filter, it seems like the general expectation is that a widget needs to perform this filter itself. So it's not the place of our rendering function to handle this.

All our widgets, plus all the default widgets are already running the `$title` through `widget_title`, so we'd be double filtering.

https://core.trac.wordpress.org/browser/trunk/src/wp-includes/widgets

AlexGStapleton:
If I remember correctly, this PR came after a user asked for more context to be provided. Is that valid or is the standard widget already doing that?

gregpriday:
Standard widgets should be providing everything we'd be providing with this filter. The only issue is that we'd rely on all widgets implementing this filter properly. They generally do if they're following the lead of the core WordPress widgets.

https://core.trac.wordpress.org/browser/trunk/src/wp-includes/widgets/class-wp-widget-pages.php#L55

PR #371, opened 2017-03-26
Row labels

braamgenis:
That is all.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -185,6 +185,10 @@
 			.row-colors( ( @i + 1 ) );
 		} .row-colors();
 
+		h3.so-row-label {
+			display: inline;
+		}
+
 		.so-row-toolbar {
 			.clearfix();
 
--- a/css/admin.less
+++ b/css/admin.less
@@ -186,7 +186,12 @@
 		} .row-colors();
 
 		h3.so-row-label {
-			display: inline;
+			display: inline-block;
+			font-size: 1em;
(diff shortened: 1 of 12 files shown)


gregpriday:
Code looks good.

The only thing we need is the ability to remove the row label. So when the user edits the label in the dialog, and removes the text, it should display "Row" as the dialog title, but the row label should be removed.

The easiest solution would be to ignore the label if its the same as the default. This also solves the issue where a user accidently clicks on the dialog title. Without entering anything, the row is labelled "Row". Ideally, we should ignore this.

gregpriday:
Also, when you edit a row, save the page, then edit the row again - the label the user entered should be added set in the edit dialog. Right now it's defaulting back to "Row."

gregpriday:
The implementation still needs a few tweaks.

1. Create a new page with a new row.
2. Edit that row, click on the dialog title. Don't make any changes, just click out of the dialog title then click done. The row is given a label "Row". Because the user didn't make any changes, the row should be unlabelled.

When you edit a row again, then you can properly clear the row label. This just happens the first time editing a row label.

There are a few other situations where similar things happen, but I imagine fixing it for that first situation will fix it in all the other situations I'm seeing it.

When entering content into the dialog title, pressing enter creates a newline. This is not the same as the behaviour on develop where pressing enter blurs the input and removes the newline. This applies to Widgets and Rows.

PR #383, opened 2017-04-12
Define  as it wasn't defined

LiljebergXYZ:
$grid was not defined and therefor the cell class attribute was not working. This should fix that issue.


gregpriday:
Very well spotted @LiljebergXYZ! Thanks.

PR #387, opened 2017-04-13
Custom Layout Directory

gregpriday:
This change allows other themes and plugins to add their own layout directories. A server can be setup using the [Layout Viewer](https://github.com/siteorigin/layout-viewer) theme.

### DIFF
--- a/inc/admin-layouts.php
+++ b/inc/admin-layouts.php
@@ -73,7 +73,7 @@ class SiteOrigin_Panels_Admin_Layouts {
 		
 		$type   = ! empty( $_REQUEST['type'] ) ? $_REQUEST['type'] : 'directory-siteorigin';
 		$search = ! empty( $_REQUEST['search'] ) ? trim( strtolower( $_REQUEST['search'] ) ) : '';
-		$page   = ! empty( $_REQUEST['page'] ) ? intval( $_REQUEST['page'] ) : 1;
+		$page_num = ! empty( $_REQUEST['page'] ) ? intval( $_REQUEST['page'] ) : 1;
 		
 		$return = array(
 			'title' => '',
@@ -108,7 +108,7 @@ class SiteOrigin_Panels_Admin_Layouts {
 			if ( ! empty( $search ) ) {
 				$query['search'] = $search;
 			}
-			$query['page'] = $page;
+			$query['page'] = $page_num;
 			
 			$directory_id = str_replace( 'directory-', '', $type );
 			$directories = $this->get_directories();
@@ -133,7 +133,7 @@ class SiteOrigin_Panels_Admin_Layouts {
 						$item['id']         = $item['slug'];
 						$item['type']       = $type;
 						
-						if( empty( $item['screenshot'] ) ) {
+						if( empty( $item['screenshot'] ) && ! empty( $item['preview'] ) ) {
 							$screenshot_url = add_query_arg( 'screenshot_preview', 1, $item['preview'] );
 							$item['screenshot'] = 'http://s.wordpress.com/mshots/v1/' . urlencode( $screenshot_url ) . '?w=700';
 						}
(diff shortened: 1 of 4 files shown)


inc/admin-layouts.php
```diff
@@ -0,0 +1,313 @@
...
+		if ( empty( $_REQUEST['_panelsnonce'] ) || ! wp_verify_nonce( $_REQUEST['_panelsnonce'], 'panels_action' ) ) {
+			wp_die();
+		}
+		
+		// Get any layouts that the current user could edit.
+		header( 'content-type: application/json' );
+		
+		$type   = ! empty( $_REQUEST['type'] ) ? $_REQUEST['type'] : 'directory-siteorigin';
+		$search = ! empty( $_REQUEST['search'] ) ? trim( strtolower( $_REQUEST['search'] ) ) : '';
+		$page   = ! empty( $_REQUEST['page'] ) ? intval( $_REQUEST['page'] ) : 1;
```
braamgenis: Small semantic issue, but I think this should rather be called `$page_num` or `$page_index` or similar. When I saw this being used below I had to return here to see what it was.

inc/admin-layouts.php
```diff
@@ -0,0 +1,313 @@
...
+			
+			if ( is_array( $response ) && $response['response']['code'] == 200 ) {
+				$results = json_decode( $response['body'], true );
+				if ( ! empty( $results ) && ! empty( $results['items'] ) ) {
+					foreach ( $results['items'] as $item ) {
+						$item['id']         = $item['slug'];
+						$item['type']       = $type;
+						
+						if( empty( $item['screenshot'] ) ) {
+							$screenshot_url = add_query_arg( 'screenshot_preview', 1, $item['preview'] );
```
braamgenis: What happens if there is no 'preview' property?

inc/admin-layouts.php
```diff
@@ -0,0 +1,313 @@
...
+			$panels_data = apply_filters( 'siteorigin_panels_prebuilt_layout', $layout, $lid );
+			
+			// Remove all the layout specific attributes
+			if ( isset( $panels_data['name'] ) ) unset( $panels_data['name'] );
+			if ( isset( $panels_data['screenshot'] ) ) unset( $panels_data['screenshot'] );
+			if ( isset( $panels_data['filename'] ) ) unset( $panels_data['filename'] );
+			
+		} elseif ( substr( $_REQUEST['type'], 0, 10 ) == 'directory-' ) {
+			$directory_id = str_replace( 'directory-', '', $_REQUEST['type'] );
+			$directories = SiteOrigin_Panels_Admin_Layouts::single()->get_directories();
```
braamgenis: `$this->get_directories();` ?

inc/admin-layouts.php
```diff
@@ -0,0 +1,313 @@
...
+	 */
+	function action_import_layout() {
+		if ( empty( $_REQUEST['_panelsnonce'] ) || ! wp_verify_nonce( $_REQUEST['_panelsnonce'], 'panels_action' ) ) {
+			wp_die();
+		}
+		
+		if ( ! empty( $_FILES['panels_import_data']['tmp_name'] ) ) {
+			header( 'content-type:application/json' );
+			$json = file_get_contents( $_FILES['panels_import_data']['tmp_name'] );
+			@unlink( $_FILES['panels_import_data']['tmp_name'] );
```
braamgenis: Should we perhaps log if any errors occur here?

js/siteorigin-panels/dialog/prebuilt.js
```diff
@@ -181,10 +181,10 @@ module.exports = panels.view.dialog.extend( {
 			page = 1;
 		}
 		if ( type === undefined ) {
-			type = 'directory';
+			type = 'directory-siteorigin';
 		}
 
-		if ( type === 'directory' && ! panelsOptions.directory_enabled ) {
+		if ( type.startsWith( 'directory-' ) && ! panelsOptions.directory_enabled ) {
```
braamgenis: `startsWith` doesn't seem very well supported yet.

inc/admin-layouts.php
gregpriday: The API should always be sending that. It just comes from the_permalink of the layout post type. I agree though, we should make it handle weird return values as much as possible.

inc/admin-layouts.php
gregpriday: Aaah yeah.

js/siteorigin-panels/dialog/prebuilt.js
gregpriday: Replaced with `match('^directory-')`

inc/admin-layouts.php
gregpriday: Yeah, better variable naming is better.

inc/admin-layouts.php
gregpriday: Yeah, that's a good idea. Will save that for a future enhancement though - https://github.com/siteorigin/siteorigin-panels/issues/388

PR #403, opened 2017-04-30
Widgets Refactor

gregpriday:
This PR uses autoloading for all basic widgets in Page Builder.

There are now 2 post loop widgets. One that uses the Widgets Bundle form, and another that's essentially the same as the old one with a custom form. There's a helper class for shared functionality across the 2 widgets.


siteorigin-panels.php
```diff
@@ -135,13 +136,19 @@ public static function is_legacy_browser(){
 	 * @param $class
 	 */
 	public static function autoloader( $class ) {
-		if ( strpos( $class, 'SiteOrigin_Panels_' ) === 0 ) {
-			$filename = strtolower( str_replace( array( 'SiteOrigin_Panels_', '_' ), array( '', '-' ), $class ) );
+		$filename = false;
+		if ( strpos( $class, 'SiteOrigin_Panels_Widgets' ) === 0 ) {
+			$filename = str_replace( 'SiteOrigin_Panels_Widgets_', '', $class );
+			$filename = preg_replace( '/([a-z])([A-Z])/', '$1-$2', $filename );
+			$filename = plugin_dir_path( __FILE__ ) . 'inc/widgets/' . strtolower( $filename ) . '.php';
```
gregpriday: Ideally, yeah... but we can't rename widget classes after they're out in the wild. That would break content.

gregpriday:
@braamgenis so the alternative from an architecture point of view is to have the Widget Bundle version of the Post Loop widget defined in the Widgets Bundle itself, and then treat the Post Loop widget in Page Builder as a sort of fallback.

We need to make sure that they're forwards compatible. So if a page is created with the Page Builder Post Loop widget, it should automatically start working with the Widgets Bundle Post Loop widget.

In an ideal world, we'd have multiple inheritance and make the Widgets Bundle Post Loop widget a child of both the Page Builder Post Loop and `SiteOrigin_Widget`, with a different class name, but because of our 2 constraints (no multiple inheritance, and forwards compatibility), we need to have 2 separate classes, with the same class name, inheriting from `WP_Widget` and `SiteOrigin_Widget` respectively.

gregpriday:
Another architecture we could use actually is having a Widgets Bundle Post Loop widget with a separate class name like `SiteOrigin_Widget_Post_Loop`. This class would be defined, but never registered as a widget.

The Page Builder Post Loop widget would then use this as a sort of helper class for rendering the form - which is the main functionality that we want from the Widgets Bundle. You reckon that's a better idea?

braamgenis:
Yeah, that sounds good. I think it's definitely worth trying to avoid the potential confusion of having two classes with the same name.

gregpriday:
@braamgenis - I implemented the new helper based on a `SiteOrigin_Widget`. This definitely feels a lot neater, and even improves how we handle the post loop with older versions of Widgets Bundle.

braamgenis:
Much better! :+1: Just one small thing I noticed. The helper widget is outputting the 'Template' select field slightly differently to the PB widget. It doesn't use the Loop Name as labels and uses the first item as the default. It's not a blocker though so can merge if you'd rather leave it for a later fix.

gregpriday:
Aaah, well spotted. I'll fix that up.

PR #410, opened 2017-05-09
Add filters inside row.

jtmcgrath:
Hey,

This is an updated version of PR #309 for the 3.0 update of Page Builder.

Example use case is to have the row's background across the full width of the screen, but the row's content wrapped in a div with a max-width.

Love the plugin, thanks for making it!

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -644,10 +644,14 @@ class SiteOrigin_Panels_Renderer {
 		if ( ! empty( $row_style_wrapper ) ) {
 			echo $row_style_wrapper;
 		}
+		
+		echo apply_filters( 'siteorigin_panels_before_cells', '', $row, $row_attributes );
 
 		foreach ( $row['cells'] as $ci => & $cell ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
This is unrelated to the commit at hand (I think it's a good idea) but I just want to chime in and say that, this:

> Example use case is to have the row's background across the full width of the screen, but the row's content wrapped in a div with a max-width.

Is already possible through the full width row layout. This change will make it possible to automate this without having to filter the default row layout or using JS/CSS though.

jtmcgrath:
Good point @Alex, it's been a while since I've tried using the different layouts.

From what I remember it wasn't easy to override the CSS which controls the layout - we use a class to control the content column width and keep it consistent throughout a site, whereas using the in-built layouts ended up requiring a fair bit of overriding CSS, especially if we wanted to use flex for the layout.

Will check out the new version to see if that's no longer a problem!

AlexGStapleton:
2.5 actually uses flex for most of the structure so things may be easier now.

jtmcgrath:
Oh nice, sounds like there's a lot for me to catch up on!

PR #417, opened 2017-05-17
vertical alignment

andrejarh:
Currently Page builder is limited to flexbox for vertical alignment.
In case of extending vertical alignment options, this patch prevents errors in css output.

### DIFF
--- a/inc/styles.php
+++ b/inc/styles.php
@@ -526,7 +526,8 @@ class SiteOrigin_Panels_Styles {
 			}
 
 			// Add in flexbox alignment to the main row element
-			if ( ! empty( $row['style']['cell_alignment'] ) ) {
+			$flexbox_align_styles = array('auto','flex-start', 'center', 'flex-end', 'stretch');
+			if ( ! empty( $row['style']['cell_alignment'] ) && in_array($row['style']['cell_alignment'], $flexbox_align_styles ) ) {
 				$css->add_row_css(
 					$post_id,
 					$ri,


andrejarh:
In our theme, I added custom options for the alignment with siteorigin_panels_row_style_fields filter. These options only require custom class, and the rest is then handled by css/javascript. 

However,  i noticed errors in the css output for this custom options.:
```
-webkit-align-items: custom_option;
align-items: custom_option;
```
This patch just prevents this behaviour by limiting css output to the actual flexbox property values.

inc/styles.php
```diff
@@ -526,7 +526,8 @@ static function filter_css_object( $css, $panels_data, $post_id, $layout ) {
 			}
 
 			// Add in flexbox alignment to the main row element
-			if ( ! empty( $row['style']['cell_alignment'] ) ) {
+			$flexbox_align_styles = array('auto','flex-start', 'center', 'flex-end', 'stretch');
+			if ( ! empty( $row['style']['cell_alignment'] ) && in_array($row['style']['cell_alignment'], $flexbox_align_styles ) ) {
```
braamgenis: I don't think we should be hard-coding the options here. Then we have to remember to add new options here and in the `row_style_fields` function. If you're not using `flexbox` then perhaps you can filter out the `cell_alignment` field entirely and add a custom `nofb_cell_alignment` field.

andrejarh: I like Flexbox and we use all the existing options, but we are adding additional (not flexbox related) options.

The problem, as I see it, is that vertical alignment is now limited to flexbox and it is not extendable beyond it. 
This was just an idea how to resolve it, but there might be better solution?

I would appreciate if you find another solution to add custom options to vertical alignment without creating errors in CSS output would be OK too.

braamgenis: The vertical alignment options are limited to flexbox because that's the layout system Page Builder is using. If you want to mix options related to flexbox layouts and non-flexbox layouts, you'd need to detect that the custom option isn't related to flexbox and therefore shouldn't be output as CSS `align-items` property. As I suggested previously, a custom field might help you do this. Another option is to use the `siteorigin_panels_css_object` filter and remove `cell_alignment` if it is set to one of your custom options. Then it won't output the CSS `align-items` property at all.

braamgenis:
Closing old conflicting PR as stale. Feel free to open a new issue for discussion.

PR #425, opened 2017-05-26
WP 4.8 Compatibility

### LINKED ISSUE #420: Compatibility with new WordPress 4.8 widgets
WordPress 4.8 (beta) introduces a few widgets with more dynamic forms. These forms all come up blank in Page Builder.

We need to try trigger actions and enqueue scripts to get these new widgets working.

The widgets are

* Video
* Audio
* Image
* Text (which replaces the old text widget with a rich text TinyMCE widget).

We need to make sure that we introduce compatibility in a backwards compatible way for the text widget.

braamgenis:
This resolves #420. Compatibility with new WP 4.8 widgets.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -857,15 +857,15 @@ class SiteOrigin_Panels_Admin {
 
 		ob_start();
 		if ( $this->is_js_widget( $the_widget ) ) {
-		?><div class="widget-content"><?php
+			?><div class="widget-content"><?php
 		}
 		$return = $the_widget->form( $instance );
 		do_action_ref_array( 'in_widget_form', array( &$the_widget, &$return, $instance ) );
 		if ( $this->is_js_widget( $the_widget ) ) {
-		?>
-		</div>
-		<input type="hidden" name="id_base" class="id_base" value="<?php echo esc_attr( $the_widget->id_base ); ?>" />
-		<?php
+			?>
+			</div>
+			<input type="hidden" name="id_base" class="id_base" value="<?php echo esc_attr( $the_widget->id_base ); ?>" />
+			<?php
 		}
 		$form = ob_get_clean();
 
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -49,12 +49,14 @@ class SiteOrigin_Panels_Admin {
 		add_action( 'load-appearance_page_so_panels_home_page', array( $this, 'add_help_tab' ), 12 );
 
 		add_action( 'customize_controls_print_footer_scripts', array( $this, 'js_templates' ) );
+		add_action( 'admin_footer-post.php', array( $this, 'js_widget_templates' ) );
+		add_action( 'admin_footer-post-new.php', array( $this, 'js_widget_templates' ) );
 
 		// Register all the admin actions
 		add_action( 'wp_ajax_so_panels_builder_content', array( $this, 'action_builder_content' ) );
 		add_action( 'wp_ajax_so_panels_widget_form', array( $this, 'action_widget_form' ) );
 		add_action( 'wp_ajax_so_panels_live_editor_preview', array( $this, 'action_live_editor_preview' ) );
-		
+
 		// Initialize the additional admin classes.
 		SiteOrigin_Panels_Admin_Widget_Dialog::single();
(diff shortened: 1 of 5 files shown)


gregpriday:
We just need to make sure this is backwards compatible with older versions of WordPress - mainly with the text widget. When using an old version of WordPress with the old non-js text widget, I'm seeing this error.

screen shot 2017-05-26 at 6 14 55 pm

inc/admin.php
```diff
@@ -754,6 +762,23 @@ function js_templates() {
 		include plugin_dir_path( __FILE__ ) . '../tpl/js-templates.php';
 	}
 
+
+	/**
+	 * Need to render templates for new WP 4.8 widgets when not on the 'widgets' screen or in the customizer.
+	 */
+	function js_widget_templates() {
+		$screen = get_current_screen();
```
gregpriday: Right now, this will be getting enqueued even on pages that don't have Page Builder active. 

Maybe `js_widget_templates` could be called inside `enqueue_admin_scripts`, after we've already checked that we're on the right page using `self::is_admin()` - https://github.com/siteorigin/siteorigin-panels/blob/develop/inc/admin.php#L227

This also means that when we have Page Builder in different places (like in Advanced Custom Fields, one day), it'd be enqueued properly.

js/siteorigin-panels/view/widgets/text-widget.js
```diff
@@ -0,0 +1,17 @@
+var $ = jQuery;
+
+var textWidget = {
+	addWidget: function( idBase, widgetContainer, widgetId ) {
+		var component = wp.textWidgets;
+
+		var widgetControl = new component.TextWidgetControl({
```
gregpriday: We need to check for the existence of `TextWidgetControl`, from my comment here - https://github.com/siteorigin/siteorigin-panels/pull/425#issuecomment-304324650

js/siteorigin-panels/view/widgets/media-widget.js
```diff
@@ -0,0 +1,37 @@
+var $ = jQuery;
+
+var mediaWidget = {
+	addWidget: function( idBase, widgetContainer, widgetId ) {
+		var component = wp.mediaWidgets;
+
+		var ControlConstructor = component.controlConstructors[ idBase ];
```
gregpriday: Shouldn't objects always start with lowercase, if we're using camelcase?

gregpriday: Aaah, nevermind, that's a constructor, so all good.

inc/admin.php
```diff
@@ -831,8 +856,17 @@ function render_form( $widget, $instance = array(), $raw = false, $widget_number
 		$the_widget->number = $widget_number;
 
 		ob_start();
+		if ( $this->is_js_widget( $the_widget ) ) {
+		?><div class="widget-content"><?php
+		}
 		$return = $the_widget->form( $instance );
 		do_action_ref_array( 'in_widget_form', array( &$the_widget, &$return, $instance ) );
+		if ( $this->is_js_widget( $the_widget ) ) {
+		?>
```
gregpriday: Would be a little easier to follow these blocks in the `if` statements were indented one level.

inc/admin.php
braamgenis: I'm not sure what you mean? `enqueue_admin_scripts` is also being called on pages that don't have PB "active". We need to have the templates available on pages so that when a widget is added it can be rendered properly. I think it would take a largish refactor to avoid enqueueing scripts etc on pages until we know PB is active.

gregpriday: Aaah, missed this before merging. Oh well, we can look at it in `develop`.

`enqueue_admin_scripts` is called on every page, but it has the `self::is_admin()` check to make sure the current page is using Page Builder (unless `$force` is true).

https://github.com/siteorigin/siteorigin-panels/blob/develop/inc/admin.php#L105

Right now `js_widget_templates` is being called on pages like WooCommerce/EDD product pages and any other custom post type. This opens us up to potential conflicts.

Plus, having all this JS widget code in `enqueue_admin_scripts` gives us the benefit of having a single function we can call to enqueue everything to make Page Builder work for situations where it's outside the context of a page. For example, the custom home page interface.

PR #440, opened 2017-06-29
Tutorials dropdown

gregpriday:
Adds a simple way for users to view our educational content inside Page Builder.

### DIFF
--- a/inc/admin-tutorials.php
+++ b/inc/admin-tutorials.php
@@ -27,14 +27,20 @@ class SiteOrigin_Panels_Admin_Tutorials {
 		
 		header( 'content-type:application/json' );
 		
-		$response = wp_remote_get('https://siteorigin.com/wp-json/siteorigin/v1/tutorials/page-builder/');
-		if ( is_array( $response ) && $response['response']['code'] == 200 ) {
-			$tutorials = json_decode( $response['body'] );
-			echo json_encode( $tutorials );
+		$tutorials = get_transient( 'siteorigin_panels_tutorials' );
+		
+		if( empty( $tutorials ) ) {
+			$response = wp_remote_get('https://siteorigin.com/wp-json/siteorigin/v1/tutorials/page-builder/');
+			if ( is_array( $response ) && $response['response']['code'] == 200 ) {
+				$tutorials = json_decode( $response['body'] );
+				set_transient( 'siteorigin_panels_tutorials', $tutorials, 86400 );
+			}
+			else {
+				$tutorials = array();
+			}
 		}
-		else {
-			echo json_encode( array() );
-		}
+		
+		echo json_encode( $tutorials );
 		
 		wp_die();
 	}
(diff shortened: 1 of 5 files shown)


inc/admin-tutorials.php
```diff
@@ -0,0 +1,42 @@
...
+		
+		header( 'content-type:application/json' );
+		
+		$response = wp_remote_get('https://siteorigin.com/wp-json/siteorigin/v1/tutorials/page-builder/');
+		if ( is_array( $response ) && $response['response']['code'] == 200 ) {
+			$tutorials = json_decode( $response['body'] );
+			echo json_encode( $tutorials );
+		}
+		else {
+			echo json_encode( array() );
```
braamgenis: In this case, perhaps return a message to tell the user either something went wrong or there aren't any tutorials available? Otherwise, it just shows the user an empty block.

inc/admin.php
```diff
@@ -275,6 +275,7 @@ function enqueue_admin_scripts( $prefix = '', $force = false ) {
 				'directory_enabled'         => ! empty( $directory_enabled ),
 				'tutorials_enabled'         => ! empty( $tutorials_enabled ),
 				'copy_content'              => siteorigin_panels_setting( 'copy-content' ),
+				'cache'						=> array(),
```
braamgenis: Shouldn't this be something like `get_transient( 'siteorigin_panels_tutorials' )` ?

tpl/js-templates.php
```diff
@@ -39,17 +39,31 @@
...
 
 			<?php if( SiteOrigin_Panels::display_premium_teaser() ) : ?>
-				<a class="so-tool-button so-learn" title="<?php echo esc_attr_e( 'Page Builder Addons', 'siteorigin-panels' ) ?>" href="<?php echo esc_url( SiteOrigin_Panels::premium_url() ) ?>" style="margin-left: 10px;">
+				<a class="so-tool-button so-premium" title="<?php echo esc_attr_e( 'Page Builder Addons', 'siteorigin-panels' ) ?>" href="<?php echo esc_url( SiteOrigin_Panels::premium_url() ) ?>" target="_blank" style="margin-left: 10px;">
 					<span class="so-panels-icon so-panels-icon-addons"></span>
 					<span class="so-button-text"><?php echo esc_html_e( 'Addons', 'siteorigin-panels' ) ?></span>
 				</a>
 			<?php endif; ?>
 				
 			<?php if( ! get_theme_support( 'siteorigin-panels' ) ) : ?>
```
gregpriday: Aaah, that's a mistake. Should have changed it to check if the user has learn links enabled in settings.

inc/admin.php
gregpriday: I was thinking that, but I actually prefer having it loaded slightly later rather. There are quite a few interfaces that use Page Builder, and having all the tutorial content always in the header might be a bit heavy.

PR #458, opened 2017-08-23
Avoid relative asset URLs

### LINKED ISSUE #447: Asset URLs are invalid and break when used with caching plugins
The urls of the scripts/styles/images that siteorigin-panels loads have "../" in them.
Example: `/wp-content/plugins/siteorigin-panels/inc/../css/front-flex.css`
This should in fact be: `/wp-content/plugins/siteorigin-panels/css/front-flex.css`

While URLs like that would normally be correctly resolved, plugins that manipulate resources don't always do that. One example is [WP Rocket](https://wp-rocket.me/) - the current version of so-panels breaks with the asset concatenation system in the current version of WP Rocket.

I think this should be fixed on the SO side - the URLs that a plugin generates should be valid without assuming that the webserver will resolve path aliases.

Relevant line in the code:
https://github.com/siteorigin/siteorigin-panels/blob/d495e25770ec09ed173622af014862fe55700b7e/inc/styles.php#L48

braamgenis:
To fix #447. Avoid using relative asset URLs which may break caching plugins.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -234,16 +234,22 @@ class SiteOrigin_Panels_Admin {
 		if ( $force || self::is_admin() ) {
 			// Media is required for row styles
 			wp_enqueue_media();
-			wp_enqueue_script( 'so-panels-admin', siteorigin_panels_url( 'js/siteorigin-panels' ) . SITEORIGIN_PANELS_VERSION_SUFFIX . SITEORIGIN_PANELS_JS_SUFFIX . '.js', array(
-				'jquery',
-				'jquery-ui-resizable',
-				'jquery-ui-sortable',
-				'jquery-ui-draggable',
-				'underscore',
-				'backbone',
-				'plupload',
-				'plupload-all'
-			), SITEORIGIN_PANELS_VERSION, true );
+			wp_enqueue_script(
+				'so-panels-admin',
+				siteorigin_panels_url( 'js/siteorigin-panels' . SITEORIGIN_PANELS_VERSION_SUFFIX . SITEORIGIN_PANELS_JS_SUFFIX . '.js' ),
+				array(
+					'jquery',
+					'jquery-ui-resizable',
+					'jquery-ui-sortable',
+					'jquery-ui-draggable',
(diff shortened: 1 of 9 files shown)


inc/admin.php
```diff
@@ -234,7 +234,7 @@ function enqueue_admin_scripts( $prefix = '', $force = false ) {
 		if ( $force || self::is_admin() ) {
 			// Media is required for row styles
 			wp_enqueue_media();
-			wp_enqueue_script( 'so-panels-admin', plugin_dir_url( __FILE__ ) . '../js/siteorigin-panels' . SITEORIGIN_PANELS_VERSION_SUFFIX . SITEORIGIN_PANELS_JS_SUFFIX . '.js', array(
+			wp_enqueue_script( 'so-panels-admin', siteorigin_panels_url( 'js/siteorigin-panels' ) . SITEORIGIN_PANELS_VERSION_SUFFIX . SITEORIGIN_PANELS_JS_SUFFIX . '.js', array(
```
gregpriday: For this, it might be slightly neater to have everything passed to the `siteorigin_panels_url` function. So:

```
siteorigin_panels_url( 'js/siteorigin-panels/' . SITEORIGIN_PANELS_VERSION_SUFFIX . SITEORIGIN_PANELS_JS_SUFFIX . '.js' )
```

inc/live-editor.php
```diff
@@ -76,21 +76,21 @@ function post_metadata( $value, $post_id, $meta_key ) {
 	function frontend_scripts() {
 		wp_enqueue_script(
 			'live-editor-front',
-			plugin_dir_url( __FILE__ ) . '../js/live-editor/live-editor-front' . SITEORIGIN_PANELS_JS_SUFFIX . '.js',
+			siteorigin_panels_url( 'js/live-editor/live-editor-front' ) . SITEORIGIN_PANELS_JS_SUFFIX . '.js',
```
gregpriday: Here especially, I think we'd want to have everything passed into `siteorigin_panels_url` if possible.

PR #464, opened 2017-09-04
Hide Cell Vertical Alignment options if Legacy Layout is set to always

AlexGStapleton:
There's no reason to show this setting if it won't do anything due to Legacy Layout always being enabled.

### DIFF
--- a/inc/styles.php
+++ b/inc/styles.php
@@ -542,7 +542,7 @@ class SiteOrigin_Panels_Styles {
 			}
 
 			// Add in flexbox alignment to the main row element
-			if ( ( siteorigin_panels_setting( 'legacy-layout' ) != 'always' || ! self::is_legacy_browser() ) && ! empty( $row['style']['cell_alignment'] ) ) {
+			if ( ( siteorigin_panels_setting( 'legacy-layout' ) != 'always' || ! SiteOrigin_Panels::is_legacy_browser() ) && ! empty( $row['style']['cell_alignment'] ) ) {
 				$css->add_row_css(
 					$post_id,
 					$ri,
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
:(
No idea why I didn't pick up on the flaws that required the two extra commits. :(

inc/styles.php
```diff
@@ -243,18 +243,20 @@ static function row_style_fields( $fields ) {
...
-			'group'    => 'layout',
-			'options'  => array(
-				'flex-start' => __( 'Top', 'siteorigin-panels' ),
-				'center'     => __( 'Center', 'siteorigin-panels' ),
-				'flex-end'   => __( 'Bottom', 'siteorigin-panels' ),
-				'stretch'    => __( 'Stretch', 'siteorigin-panels' ),
-			),
-			'priority' => 17,
-		);
+		if ( siteorigin_panels_setting( 'legacy-layout' ) != 'always'  ) {
```
braamgenis: Wouldn't you want to do the `SiteOrigin_Panels::is_legacy_browser()` check here too?

AlexGStapleton: No. This is what the person editing the page will see (it's when editing the row styles after all) rather than what the person on the frontend will see. The editor may have a really old browser on one computer and modern browser on another so to prevent possible issues when editing the page, I didn't check for this. That and it doesn't matter what the editor's browser can do - only the browser viewing the page on the frontend is what matters. 

Does that make sense? With that said, I can change it if you want but I don't think it's needed.

braamgenis: Ah yeah, of course. Ok, all good. :+1:

PR #473, opened 2017-09-21
Add filter to the background-image url in the CSS output

KZeni:
I personally want as much as possible to be sent via a CDN when it comes to site images & other assets. Adding the `siteorigin_panels_background_image_url` filter to the background-image URL would allow me (and others) to replace the URL with the CDN's URL. This can also be useful for those simply using Jetpack's Photon feature as the filter can allow jetpack_photon_url(); to replace the URL being used with the image on WordPress' CDN.


braamgenis:
Closing old conflicting PR as stale. Feel free to open a new issue for discussion.

PR #478, opened 2017-10-19
fixes js error for deprecated .load event and replace with .on

Madisafari:
on some themes(depending on jquery version) .load event causes a javascript error, here is the javascript error: 
`Uncaught TypeError: a.indexOf is not a function`
and this doesn't let the page builder to render the page.
.load event has been replaced by .on(...

additional info:
See bug [#11733](http://bugs.jquery.com/ticket/11733), which documents this deprecation:

> The .load() method is an ambiguous signature, it can either be an ajax load or attach/fire a "load" event. CCAO cannot tell them apart since it's a dynamic decision based on arguments.

### DIFF
--- a/js/styling.js
+++ b/js/styling.js
@@ -50,7 +50,7 @@ jQuery( function ( $ ) {
 			$( window ).trigger( 'panelsStretchRows' );
 		}
 	}
-	$( window ).resize( stretchFullWidthRows ).on("load",function(){
+	$( window ).resize( stretchFullWidthRows ).on('load',function(){
 		stretchFullWidthRows();
 	});
 	stretchFullWidthRows();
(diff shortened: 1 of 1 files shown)


js/styling.js
```diff
@@ -50,7 +50,9 @@ jQuery( function ( $ ) {
 			$( window ).trigger( 'panelsStretchRows' );
 		}
 	}
-	$( window ).resize( stretchFullWidthRows ).load( stretchFullWidthRows );
+	$( window ).resize( stretchFullWidthRows ).on('load',function(){
```
braamgenis: No need to use a closure here. It's fine to just call `stretchFullWidthRows` directly.
In fact, this could be shortened to just `$( window ).on( 'resize load', stretchFullWidthRows );`

Madisafari: @braamgenis initially I wrote it like that but Codacy gave an error "on check" and I changed it to the way codacy asked to satisfy the check.
I'll shorten the code and commit again.
thanks.

braamgenis: Well Codacy seems to be happy with it now. :+1:

PR #479, opened 2017-10-21
Added way of loading prebuilt layouts through JSON

gregpriday:
We need to decide if this is the best way for themes to include JSON layouts. Basically, the function gives themes a way of adding a prebuilt layout from a JSON file using the standard prebuilt layouts filter.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -905,6 +905,13 @@
 				padding-left: 26px;
 				background-position: left center;
 			}
+
+			.dashicons-warning {
+				color: #aa0000;
+				vertical-align: middle;
+				margin-right: 7px;
+				margin-top: -1px;
+			}
 		}
 
 		.so-buttons {
--- a/css/admin.less
+++ b/css/admin.less
@@ -1482,10 +1482,13 @@
 				font-weight: bold;
 			}
 
+			.so-directory-items-wrapper {
+				display: flex;
+				flex-flow: row wrap;
+			}
+
 			// For a list of directory items
 			.so-directory-items {
-
-				.clearfix();
 
 				.so-no-results {
 					margin: 20px 0;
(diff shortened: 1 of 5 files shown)


braamgenis:
Ah wait, so would themes call `load_layout` and then filter that loaded layout data into the layouts collection using `siteorigin_panels_prebuilt_layouts`?

braamgenis:
Just thinking out loud: What about using a convention-based approach. So PB would just load any JSON files it finds in a `layouts` directory in the active theme? Makes it a little less work for theme devs and then basically any non-technical users could also just build a layout, export it, and drop it in the aforementioned theme directory.

gregpriday:
So I really like that idea, but we just need to give themes a way of providing the name and screenshot of the design. That's not currently stored in the JSON.

Maybe what we could do is have the Page Title always stored in the JSON file and use that, then just expect a correspondingly named PNG or JPG file in the same folder as the screenshot.

We'd also need to give plugins a way of registering a folder.

If you keen on all that, I'll get to implementing.

braamgenis:
Yeah, if I understand correctly, that sounds good. :+1: The _Page Title_ being the title of the page from which the layout was exported? And that would be set by the layout export code? Then that might change the layouts JSON slightly and the layout import code would need to change to handle both layouts with and without a title.

And yeah, then I guess also add a filter for plugins to add a folder and also for theme devs who might want to customize where they keep layouts.

SwiftThemes:
Having the layouts in a folder is neat. 
Some theme might already use the layouts folder in their code structure. It's safe to name space the folder. sopb-layouts or pb-layouts may be.

Also, it would be good if we can list the requirements. Like if premium widgets or widgets from a different plugin are required .

braamgenis:
@SwiftThemes Good idea to make sure the folder won't conflict with existing folders, thanks!

As for listing requirements, we're going to try keep this simple for now. We do already have filters for the form and rendering of missing widgets. So `siteorigin_panels_missing_widget` is called when rendering the front end:

https://github.com/siteorigin/siteorigin-panels/blob/27ce7df84389872b235034c29e55dc1f89bbf129/inc/renderer.php#L470

 and `siteorigin_panels_missing_widget_form` is called when rendering the admin form:

https://github.com/siteorigin/siteorigin-panels/blob/6962a5a5cb2b29b455c2b26883ae9f13f90ac027/inc/admin.php#L861

You can filter these and modify them to let users know which plugins they need to install.

PR #484, opened 2017-11-06
Release 2.5.14

braamgenis:
* Content Cache: Add Enqueue hook to allow 3rd parties to enqueue cache friendly assets.
* Added `raw_panels_data` flag for layout imports.
* Save `ratio` and `ratio_direction` as row attributes.
* Add `rel="noopener noreferrer"` for all 3rd party/unknown links.

### DIFF
--- a/inc/post-content-filters.php
+++ b/inc/post-content-filters.php
@@ -38,10 +38,10 @@ class SiteOrigin_Panels_Post_Content_Filters {
 			$attributes[ 'data-style' ] = json_encode( $row['style'] );
 		}
 		if( ! empty( $row['ratio'] ) ) {
-			$attributes[ 'data-ratio' ] = intval( $row['ratio'] );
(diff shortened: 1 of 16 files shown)


inc/post-content-filters.php
```diff
@@ -37,6 +37,12 @@ public static function row_attributes( $attributes, $row ){
 		if( ! empty( $row['style'] ) ) {
 			$attributes[ 'data-style' ] = json_encode( $row['style'] );
 		}
+		if( ! empty( $row['ratio'] ) ) {
+			$attributes[ 'data-ratio' ] = intval( $row['ratio'] );
```
gregpriday: Wait, missed this. Can't ratio also be a float?

braamgenis: Ah, good catch! :+1:

PR #490, opened 2017-11-20
Custom paddings on full width stretched row

andrejarh:
No need to set padding to '0'. It is better to unset it with ''. 
If they are set to 0, custom row padding don't work on Full Width Stretched row. 
Setting them to empty, allows for custom row paddings to kick in, if they are set.

### DIFF
--- a/js/styling.js
+++ b/js/styling.js
@@ -17,8 +17,8 @@ jQuery( function ( $ ) {
 			$$.css( {
 				'margin-left': 0,
 				'margin-right': 0,
-				'padding-left': 0,
-				'padding-right': 0
+				'padding-left': '',
+				'padding-right': ''
 			} );
 
 			var leftSpace = $$.offset().left - fullContainer.offset().left,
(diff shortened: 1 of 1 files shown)


andrejarh:
Asking for review. 

I think it would be great to have custom row (left and right) paddings working on Full width stretched row.

To get standard Full width row custom paddings to work, it would be a bit more complicated as we would need to get the value of custom paddings inside js (this could be done with data attributes maybe), but it might also be an overkill? So I suggest to leave the calculation for standard Full width rows as it is. However there is no reason not to allow custom row paddings on Full width stretched row. 

Default side paddings (padding: 0) are better placed in CSS, then in js anyway.

@gregpriday and @Madisafari what do you think?

braamgenis:
Hi @andrejarh Thanks for the contribution! While I agree that this change is probably the better way to handle things, changing this behaviour will potentially change existing layouts for a lot of users. So to avoid a large number of users having issues we're going to leave things as they are until we’re able to find a solution that works for both existing users and users who want this slightly altered behaviour.

andrejarh:
Hi @braamgenis,
Thanks for reply.
I tested this solution with couple of layouts and they did not break.
But I understand your concern. Modifying any defaults can influence the existing users. Luckily there is a simple way to avoid this completely.

Now default left and right row paddings are set to 0 with javascript. 
These defaults can easily be added to CSS (it  is where they should be anyway): 
`.panel-row-style {
	padding-left: 0;
	padding-right: 0;
}`

In this case, defaults stay as they were, we just move them to CSS. This way the behaviour for any existing users will not be changed, unless they added the paddings to it manually in the row settings. In which case they probably wanted to add the padding anyway. 

What do you think?

ellice909:
+1 to adrejarh's idea.

I am trying to edit the JS to take out the 0 padding. It really took out all convenience of the in-page editor, and now I will have to be careful of any future plug-in updates. I have been dealing with this issue for a year or two. It typically occurs on my clients' sites but now it will be on my site as well. It looks like it as been some months with no action on this front. I'd be eager to see this change.

braamgenis:
This has been resolved in 657b143a1122004226cfaaca3fd2b439f4d97a34 in a way that avoids potential issues for existing users.

PR #495, opened 2017-12-04
Inline styles filter

### LINKED ISSUE #466: Add the ability to disable inline styling CSS
Before SiteOrigin Page Builder 2.5, you were able to disable the SiteOrigin Page Builder Inline CSS by simply removing its filters ([reference](https://siteorigin.com/thread/disable-inline-styling-of-the-layout/)).After 2.5 refactor, it's no longer possible to do this.

braamgenis:
To resolve #466 This adds a filter to allow third party developers to modify or remove inline styles.

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -22,6 +22,7 @@ class SiteOrigin_Panels_Renderer {
 	 */
 	public function add_inline_css( $post_id, $css ) {
 		if ( is_null( $this->inline_css ) ) {
+			// Initialize the inline CSS array and add actions to handle printing.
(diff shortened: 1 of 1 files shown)


gregpriday:
@braamgenis this looks good, just double checking, does it work in the same way as the pre 2.5 filter?

braamgenis:
@gregpriday No, it's a new filter altogether but the previous method of disabling was still there just slightly different. See [this comment](https://github.com/siteorigin/siteorigin-panels/issues/466#issuecomment-348974126) on the issue.

Issue #500, opened 2017-12-20
Notice when unregistering 'SiteOrigin_Panels_Widgets_Layout' widget


mariovalney:
If you unregister 'SiteOrigin_Panels_Widgets_Layout' widget a notice will be displayed:

> PHP Notice:  Undefined index: title in .../wp-content/plugins/siteorigin-panels/inc/admin-widget-dialog.php on line 79

AlexGStapleton:
Unfortunately, it appears as though you've submitted a PR for the fork rather than the upstream. I would recommend giving [this article](https://help.github.com/articles/creating-a-pull-request-from-a-fork/) a quick read.

mariovalney:
You are right... hahah redone.

PR #501, opened 2017-12-26
Checking widget to avoid notice

mariovalney:
Checking 'SiteOrigin_Panels_Widgets_Layout' to avoid notice when unregistering the widget.

Issue #500


braamgenis:
Thanks for the contribution! 👍

PR #504, opened 2018-01-10
Fix/delete by meta key

igmoweb:
In order to play fair WordPress meta cache. It would be better if `siteorigin_panels_cache` meta key is deleted by using `delete_post_meta_by_key` function instead of direct queries to DB.


braamgenis:
Hi @igmoweb Thanks for the contribution. 👍

PR #510, opened 2018-02-12
Mutliple editors preview fix

### LINKED ISSUE #505: Preview bugged for post creator when there is an autosave from another user
If there are multiple users working on a post and one of them creates an autosave (and then for example closes the window so the lock expires), the other user (not sure if it is always the owner of the post) will always be presented with an autosave of the other user.

**How to reproduce:**
1) Create a post with user A, add site origin editor and add some content
2) Save draft, Log out, delete lock from DB in wp_postmeta (emulates user closing the window and lock expiring)
3) Log in as user B
4) Edit the post changing some of the contents, click Preview to generate an autosave
5) Log out, delete lock from DB
6) Log in as user A, edit the post
7) Change the contents
8) Click preview and you should be presented with the user B's version of the autosave

Even if you save the post as draft and then preview it, you will again be presented with B's autosave

**Possible solution:**
https://github.com/siteorigin/siteorigin-panels/blob/develop/siteorigin-panels.php#L279
```diff
		if ( $preview ) {
-			$preview_post = wp_get_post_autosave( $post_id );
+			$preview_post = wp_get_post_autosave( $post_id, get_current_user_id() );
			if ( ! empty( $preview_post ) ) {
				$post_id = $preview_post->ID;
			}
		}
```

Because there is a check in `wp_get_post_autosave` whether the autosave belongs to the user previewing the post, it will retrieve the first autosave entry no matter if it belongs to the current user or not. This post is then used inside the `generate_post_content` function and the content is generated from the autosave instead of the latest version

Sharsie:
This should fix #505, I've been using it for a while, however I am not 100% sure that this is all that needs to be fixed.

I haven't come accross any other use case though which would break the preview.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -276,7 +276,7 @@ class SiteOrigin_Panels {
 		$post_id = get_the_ID();
 		// If we're viewing a preview make sure we load and render the autosave post's meta.
 		if ( $preview ) {
-			$preview_post = wp_get_post_autosave( $post_id );
+			$preview_post = wp_get_post_autosave( $post_id, get_current_user_id() );
 			if ( ! empty( $preview_post ) ) {
 				$post_id = $preview_post->ID;
 			}


braamgenis:
Hi @Sharsie Thanks for the contribution! Looks good. 👍

Issue #367, opened 2017-03-22
Remove styles in header and footer and include them in a stylesheet? 


LittleHamster:
Thanks to @gregpriday the inline style attributes are now fixed. 
I was wondering if you guys can look into this issue as well, so I created a couple of rows and columns and this is added to the header: 
```
<style type="text/css" media="all" id="siteorigin-panels-grids-wp_head">
/* Layout 408 */ #pgc-408-0-0 , #pgc-408-0-1 { width:50% } #pg-408-0 .panel-grid-cell { float:left } #pl-408 .panel-grid-cell .so-panel { margin-bottom:30px } #pl-408 .panel-grid-cell .so-panel:last-child { margin-bottom:0px } #pg-408-0 { margin-left:-15px;margin-right:-15px } #pg-408-0 .panel-grid-cell { padding-left:15px;padding-right:15px } @media (max-width:780px){ #pg-408-0 .panel-grid-cell { float:none;width:auto } #pgc-408-0-0 { margin-bottom:30px } #pl-408 .panel-grid , #pl-408 .panel-grid-cell {  } #pl-408 .panel-grid .panel-grid-cell-empty { display:none } #pl-408 .panel-grid .panel-grid-cell-mobile-last { margin-bottom:0px }  } 
</style>
```

there is also one in the footer: 
```
<style type="text/css" media="all" id="siteorigin-panels-grids-wp_footer">
```

should these be loaded via enqueue style in a separate css file? isn't that more efficient?

gregpriday:
Yes and no. The CSS generated by Page Builder is very specific to the page. External stylesheets are great when there's a lot of shared styling. So as your user browses your site, they always have the same stylesheet cached.

For single-use CSS like Page Builder generates, it's actually more efficient to have it inline because it reduces the need for for an additional HTTP request.

That said, we would like to add the option of having external stylesheets for people who want it, but it's not necessarily the best/only solution.

There shouldn't be styling in both your header and footer though. Assuming you're viewing a single page. I'll take a look at this.

LittleHamster:
@gregpriday thanks for taking a look at this, it would be great to have an option to include all of these (all styles for different pages, posts) in a single style.css file.

Rajik:
Any updates on this enhancement? We are in need of this feature and it would be great to know the timeline for this.

gregpriday:
[...]

PR #515, opened 2018-03-19
Render CSS in HTML as inline or in a new file - toggle with a setting

Rajik:
This PR lets the admin choose whether they want to have CSS in a <style> tag in the HTML of the page or have a separate CSS file. This PR addresses [this](https://github.com/siteorigin/siteorigin-panels/issues/367) issue/feature request. 
The setting is set as inline as default to make it compatible with the existing distributions of the plugin.

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -1,5 +1,4 @@
 <?php
-
 class SiteOrigin_Panels_Renderer {
 
 	private $inline_css;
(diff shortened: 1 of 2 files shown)


Rajik:
@gregpriday Requesting a code review.

gregpriday:
Sorry for the late reply on this @Rajik. Unfortunately there are a few issues with this PR. Most importantly, it seems to be using `file_put_contents` for file writes, which isn't best practice for security reasons. You should be using the WordPress [Filesystem API](https://codex.wordpress.org/Filesystem_API).

It's also writing files on every page request, which will be slow and saving those files to a folder inside the plugin, rather than into a folder in `wp-content`.

I'd prefer that any PR added to address #367 just adds hooks and filters to make having separate CSS files possible.

Issue #142, opened 2016-01-18
Feature request: Adding a new widget should open the controls for that widget immediately


richardtape:
Right now when you add a widget to the editor you then have to click the edit button to set the properties for it. I know, right? A whole extra click. Facetiousness aside, if you have many widgets in the editor it can be a little awkward for the user to grok exactly which widget they've just added. It'd be nice to add the widget and immediately open the editing screen for it.

Thoughts?

gregpriday:
As far as I remember, this is how we used to do it before the version 2.0 update. Perhaps readding it as an optional setting for users who prefer it is a good idea though.

It was a gut feel that it might feel a bit jarring having the widget window open up right away. Also, because of the nice widget navigation we added (the option to go left/right through widgets on a page), it felt more natural to let a user quickly setup their page and widgets first, then go through and add their content. Again, just a gut feel.

richardtape:
Thanks for replying @gregpriday 

Interesting. Even if the feature was still there but behind a filter, I could get behind that. Decisions, not options. (But with a bit of flexibility should you want it... :) )

I'm happy to see if I can re-add this if you could point me to something in the history about where this was implemented?

jambox:
Agree with @richardtape here

PR #540, opened 2018-06-05
Implement instant open feature

gregpriday:
Resolves https://github.com/siteorigin/siteorigin-panels/issues/142

### DIFF
--- a/js/siteorigin-panels/dialog/widgets.js
+++ b/js/siteorigin-panels/dialog/widgets.js
@@ -185,6 +185,8 @@ module.exports = panels.view.dialog.extend( {
 
 		this.closeDialog();
 		this.builder.model.refreshPanelsData();
+
+		widget.trigger('user_created_new');
 	},
 
 	/**
--- a/js/siteorigin-panels/dialog/widgets.js
+++ b/js/siteorigin-panels/dialog/widgets.js
@@ -171,6 +171,7 @@ module.exports = panels.view.dialog.extend( {
 	 */
 	widgetClickHandler: function ( e ) {
 		// Add the history entry
+		this.builder.trigger('before_user_adds_widget');
 		this.builder.addHistoryEntry( 'widget_added' );
 
 		var $w = $( e.currentTarget );
@@ -186,7 +187,7 @@ module.exports = panels.view.dialog.extend( {
 		this.closeDialog();
 		this.builder.model.refreshPanelsData();
 
-		widget.trigger('user_created_new');
+		this.builder.trigger('after_user_adds_widget', widget);
 	},
 
 	/**
--- a/js/siteorigin-panels/view/widget.js
+++ b/js/siteorigin-panels/view/widget.js
@@ -24,9 +24,7 @@ module.exports = Backbone.View.extend( {
 		this.model.on( 'user_edit', this.editHandler, this );
 		// When a user wants to duplicate the widget model
 		this.model.on( 'user_duplicate', this.duplicateHandler, this );
-		// When the user created a new widget through a direct action
-		this.model.on( 'user_created_new', this.createdNewHandler, this );
-		
+
 		this.model.on( 'destroy', this.onModelDestroy, this );
 		this.model.on( 'visual_destroy', this.visualDestroyModel, this );
 
@@ -81,6 +79,10 @@ module.exports = Backbone.View.extend( {
 			// Setup the dialog to load the form
 			dialog.setupDialog();
 		}
+
+		// Add the global builder listeners
+		this.listenTo(this.cell.row.builder, 'after_user_adds_widget', this.afterUserAddsWidgetHandler);
+
 
 		return this;
 	},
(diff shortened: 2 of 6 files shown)


js/siteorigin-panels/view/widget.js
```diff
@@ -20,8 +20,13 @@ module.exports = Backbone.View.extend( {
 	 * Initialize the widget
 	 */
 	initialize: function () {
-		this.model.on( 'user_edit', this.editHandler, this );				 // When a user wants to edit the widget model
-		this.model.on( 'user_duplicate', this.duplicateHandler, this );	   // When a user wants to duplicate the widget model
+		// When a user wants to edit the widget model
+		this.model.on( 'user_edit', this.editHandler, this );
```
braamgenis: It's better to use `this.listenTo( this.model, 'user_edit', this.editHandler );` as it allows all listeners to be removed more easily and removes the need for providing `this` as context every time.

js/siteorigin-panels/view/widget.js
```diff
@@ -20,8 +20,13 @@ module.exports = Backbone.View.extend( {
...
+		// When a user wants to duplicate the widget model
+		this.model.on( 'user_duplicate', this.duplicateHandler, this );
+		// When the user created a new widget through a direct action
+		this.model.on( 'user_created_new', this.createdNewHandler, this );
```
braamgenis: We should be consistent with tenses when triggering events. E.g. 'user_duplicate' should be triggered just before the duplication event is about to take place, where `user_created_new` should be triggered after the event has taken place.

** I just looked a bit closer at the code and realized these events are being used differently, which is confusing. `user_edit` and `user_duplicate` are actually user intentions? If so they should be triggered on views, instead of models... I think. 😕

js/siteorigin-panels/dialog/widgets.js
```diff
@@ -185,6 +185,8 @@ module.exports = panels.view.dialog.extend( {
 
 		this.closeDialog();
 		this.builder.model.refreshPanelsData();
+
+		widget.trigger('user_created_new');
```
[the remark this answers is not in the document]
gregpriday: The main difference is that `widget_added` for history triggers before adding the widget, and we needed something that'd trigger after the widget is created. If we wanted to get things really need, I agree going with a consistent `this.builder.model.trigger('widget_added', widget )` and letting all the views hook into that would be way better. Definitely keen to hit that at some point.

The only reason we couldn't really trigger this inside the model is we needed some way of differentiating between user events and more automatic events. For example, has the user added a widget (in which case we need a history entry and possibly instant open the dialog), or is the widget being added as a result of loading `panels_data`. Sure there's a much smarter way to do this than we're currently doing though.

js/siteorigin-panels/view/widget.js
gregpriday: Awesome, will change that.

js/siteorigin-panels/view/widget.js
gregpriday: Agree with keeping tense the same. Will change that.

It's been a while, so I can't quite remember why we're triggering actions on the models. I know it had to do with a combination of factors like:

* Only having access to the model, and not the view in certain instances
* Needing to know the difference between user triggered and automatic events
* Overall inexperience with Backbone when I wrote some of this code :P

Will take a deep dive and see how much we can rearchitect this code to make it more logical and consistent. We can save that as a little project for the 2.7 update.

gregpriday:
@braamgenis I'll make some of the more straight forward changes, and save the more fundamental changes for a bigger 2.7 update.

js/siteorigin-panels/dialog/widgets.js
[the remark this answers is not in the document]
braamgenis: I think we shouldn't be triggering those events at all when loading `panels_data`, as that's more of an internal process. I think we could have something like 
`function addWidget( widget, silent )` with `silent` set to `true` when loading `panels_data`.

js/siteorigin-panels/view/widget.js
```diff
@@ -278,6 +283,18 @@ module.exports = Backbone.View.extend( {
...
 		this.cell.buildContextualMenu( e, menu );
+	},
+
+	/**
+	 * Handler for any action after the user adds a new widget.
+	 * @param widget
+	 */
+	afterUserAddsWidgetHandler: function( widget ) {
+		if( this.model === widget && panelsOptions.instant_open ) {
+			setTimeout(function(){
```
braamgenis: Ok, last thing. 😛 Is`setTimeout` really necessary? Any way we could avoid using it and use events and/or callbacks somehow?

gregpriday: One thing I did change (and maybe this is what you meant) was using `setTimeout(this.editHandler, 350);`. Much better.

braamgenis: Ah ok, all good then.

Issue #541, opened 2018-06-05
Trigger user driven events on views, rather than models


gregpriday:
Based on the discussion here: https://github.com/siteorigin/siteorigin-panels/pull/540

There are several cases where it makes more sense to trigger events on the views that take a model as an argument, rather than directly on the models. For simplicity, these events could all be triggered on the main builder view, but possibly also view being interacted with.

Examples of actions that we should move are user adding, removing, duplicating rows and widgets.

gregpriday:
Priority too low. Must increase.

gregpriday:
Implemented here - https://github.com/siteorigin/siteorigin-panels/tree/feature/view-based-events. Will merge after https://github.com/siteorigin/siteorigin-panels/pull/540.

PR #542, opened 2018-06-06
View based events

gregpriday:
Resolves https://github.com/siteorigin/siteorigin-panels/issues/541

This PR takes out most/all of the cases where we were triggering events on models that should have been triggered on views. In the future I'd like to have most user events trigger a view action somewhere, but we can tackle that when we actually need it.

### DIFF
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -20,6 +20,7 @@ module.exports = panels.view.dialog.extend({
 		'click .row-set-form button.set-row': 'setCellsFromForm',
 	},
 
+	rowView: null,
 	dialogIcon: 'add-row',
 	dialogClass: 'so-panels-dialog-row-edit',
 	styleType: 'row',
@@ -749,7 +750,7 @@ module.exports = panels.view.dialog.extend({
 	 */
 	deleteHandler: function () {
 		// Trigger a destroy on the model that will happen with a visual indication to the user
-		this.visualDestroyModel();
+		this.rowView.visualDestroyModel();
 		this.closeDialog({silent: true});
 
 		return false;
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -749,7 +749,7 @@ module.exports = panels.view.dialog.extend({
 	 */
 	deleteHandler: function () {
 		// Trigger a destroy on the model that will happen with a visual indication to the user
-		this.model.trigger('visual_destroy');
+		this.visualDestroyModel();
 		this.closeDialog({silent: true});
 
 		return false;
--- a/js/siteorigin-panels/dialog/widget.js
+++ b/js/siteorigin-panels/dialog/widget.js
@@ -102,11 +102,12 @@ module.exports = panels.view.dialog.extend( {
 		if ( widgets.length <= 1 ) {
 			return false;
 		}
-		var currentIndex = widgets.index( this.widgetView.$el ), widgetView;
+		var currentIndex = widgets.index( this.widgetView.$el );
 
 		if ( currentIndex === 0 ) {
 			return false;
 		} else {
+			var widgetView;
 			do {
 				widgetView = widgets.eq( --currentIndex ).data( 'view' );
 				if ( ! _.isUndefined( widgetView ) && ! widgetView.model.get( 'read_only' ) ) {
@@ -128,11 +129,12 @@ module.exports = panels.view.dialog.extend( {
 			return false;
 		}
 
-		var currentIndex = widgets.index( this.widgetView.$el ), widgetView;
+		var currentIndex = widgets.index( this.widgetView.$el );
 
 		if ( currentIndex === widgets.length - 1 ) {
 			return false;
 		} else {
+			var widgetView;
 			do {
 				widgetView = widgets.eq( ++currentIndex ).data( 'view' );
 				if ( ! _.isUndefined( widgetView ) && ! widgetView.model.get( 'read_only' ) ) {
--- a/js/siteorigin-panels/dialog/widget.js
+++ b/js/siteorigin-panels/dialog/widget.js
@@ -102,7 +102,7 @@ module.exports = panels.view.dialog.extend( {
 		if ( widgets.length <= 1 ) {
 			return false;
 		}
-		var currentIndex = widgets.index( this.widgetView.$el );
+		var currentIndex = widgets.index( this.widgetView.$el ), widgetView;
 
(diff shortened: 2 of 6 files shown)


js/siteorigin-panels/dialog/row.js
```diff
@@ -749,7 +749,7 @@ module.exports = panels.view.dialog.extend({
 	 */
 	deleteHandler: function () {
 		// Trigger a destroy on the model that will happen with a visual indication to the user
-		this.model.trigger('visual_destroy');
+		this.visualDestroyModel();
```
braamgenis: I don't see a `visualDestroyModel` function anywhere in this view? I'm guessing it's meant to call the `visualDestroyModel` function in the row view. Regardless this should just be calling `this.model.destroy()` and any other actions that need to happen should listen for the `destroy` event on the model.

js/siteorigin-panels/dialog/widget.js
```diff
@@ -274,16 +274,16 @@ module.exports = panels.view.dialog.extend( {
 	 * @returns {boolean}
 	 */
 	deleteHandler: function () {
-
-		this.model.trigger( 'visual_destroy' );
+		this.widgetView.visualDestroyModel();
```
braamgenis: Same here. Should just be calling `this.model.destroy()`

js/siteorigin-panels/dialog/widget.js
```diff
@@ -274,16 +274,16 @@ module.exports = panels.view.dialog.extend( {
...
 		this.closeDialog( {silent: true} );
 		this.builder.model.refreshPanelsData();
 
 		return false;
 	},
 
 	duplicateHandler: function () {
-		this.model.trigger( 'user_duplicate' );
+		// Call the widget duplicate handler directly
+		this.widgetView.duplicateHandler();
```
braamgenis: Views shouldn't be reaching outside themselves and calling functions on other views, but yeah, then you'd have to duplicate the duplication handling code here. 😅 I guess this is one of the situations where a controller class would come in handy.

js/siteorigin-panels/view/builder.js
```diff
@@ -88,21 +88,21 @@ module.exports = Backbone.View.extend( {
 		} );
 		
 		// When the data changes in the model, store it in the field
-		this.model.on( 'change:data load_panels_data', this.storeModelData, this );
+		this.listenTo( this.model, 'change:data load_panels_data', this.storeModelData );
+		this.listenTo( this.model, 'change:data load_panels_data', this.toggleWelcomeDisplay );
 		
 		// Handle a content change
 		this.on( 'content_change', this.handleContentChange, this );
```
braamgenis: These can also use `this.listenTo`.

js/siteorigin-panels/view/row.js
```diff
@@ -21,10 +21,9 @@ module.exports = Backbone.View.extend( {
 		var rowCells = this.model.get('cells');
 		rowCells.on( 'add', this.handleCellAdd, this );
 		rowCells.on( 'remove', this.handleCellRemove, this );
```
braamgenis: Can also use `this.listenTo` here.

js/siteorigin-panels/view/row.js
```diff
@@ -21,10 +21,9 @@ module.exports = Backbone.View.extend( {
...
-		this.model.on( 'reweight_cells', this.resize, this );
 
-		this.model.on( 'destroy', this.onModelDestroy, this );
-		this.model.on( 'visual_destroy', this.visualDestroyModel, this );
+		this.listenTo( this.model, 'reweight_cells', this.resize );
+		this.listenTo( this.mode, 'destroy', this.onModelDestroy );
```
braamgenis: Typo. `model` missing the `l`

js/siteorigin-panels/view/widget.js
```diff
@@ -186,13 +178,12 @@ module.exports = Backbone.View.extend( {
...
-			thisView.cell.row.resize();
-			thisView.model.destroy();
-			thisView.cell.row.builder.model.refreshPanelsData();
-			thisView.remove();
-		} );
+			this.cell.row.resize();
+			this.model.destroy();
+			this.cell.row.builder.model.refreshPanelsData();
+			this.remove();
+		}.bind(this) );
```
braamgenis: Noice.

js/siteorigin-panels/dialog/row.js
gregpriday: You're right. That should be getting called on the row view, not the dialog view. Will fix.

`visualDestroyModel` destroys the model with a fade out as a visual cue, there are other times when the model is destroyed and we don't want a visual cue. Like when the row is removed and we need to remove all its child widgets. Using a view function was a work around for all this.

If we can find a way to pass an argument to `model.destroy()` that gets passed to the view, then we can avoid `visualDestroyModel`.

js/siteorigin-panels/dialog/widget.js
gregpriday: Possible we can just call the `this.widgetView.duplicateHandler` directly from events in the dialog. Not sure if that's better practice?

js/siteorigin-panels/view/builder.js
gregpriday: Can something use `listenTo` when listening to itself? The function is `object.listenTo(other, event, callback)` and the description starts with:

>Tell an object to listen to a particular event on an other object.

So shouldn't this still be using `on`?

js/siteorigin-panels/view/row.js
gregpriday: Same case here. Isn't `listenTo` only for listening to events on other objects?

js/siteorigin-panels/view/row.js
gregpriday: Good eye 👍

gregpriday: OR SHOULD I SAY, GOOD L?! No? Okay...

js/siteorigin-panels/view/widget.js
gregpriday: Approval feels good.

js/siteorigin-panels/view/builder.js
braamgenis: Ah, yeah, I think you're right. Using `this.listenTo` on `this` will either not work or just have the same effect as calling `this.on`.

braamgenis:
I think this horse is dead. 😛 Merging.

js/siteorigin-panels/view/row.js
gregpriday: My bad, rad, will fix that in develop.

PR #543, opened 2018-06-07
Add options to change get hash by widget number

dvekotletki:
Hello! We got a problem with using Page Builder in batch with Widgets bundle.
When widget has styles, it generates them once a week when the page is requested.
On this way, we got some middleware before user get the page. We have FCGI and CDN, those two tools cache pages and their static to make it faster for end user.
We trying to figure out how we could fix disappearing widget styles and that what we came to:

We add an option to make `*.css` file name generated with not unique hash everytime, but with sliced md5 of widget-id hash. You can check it here: `inc/settings.php:121`, `inc/settings.php:257`

### DIFF
--- a/inc/settings.php
+++ b/inc/settings.php
@@ -118,6 +118,7 @@ class SiteOrigin_Panels_Settings {
 		// The general fields
 		$defaults['post-types']             = array( 'page', 'post' );
 		$defaults['live-editor-quick-link'] = true;
+		$defaults['hash-by-number'] 		= false;
 		$defaults['admin-widget-count']     = false;
 		$defaults['parallax-motion']        = '';
 		$defaults['sidebars-emulator']      = true;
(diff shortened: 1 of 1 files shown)


dvekotletki:
Request with the same purposes you could find in so-widget-bundle here:
https://github.com/siteorigin/so-widgets-bundle/pull/712

If you have any questions or comments please write down here.
Thanks in advance.

Misplon:
Hi @dvekotletki Thanks for posting. Would it not be possible to add a row class or a widget class via the options available in Page Builder? Or to use body classes? Can you perhaps share a link to the page concerned and point out the widget/s being targetted so we can take a look at the use case?

gregpriday:
I've replied here https://github.com/siteorigin/so-widgets-bundle/pull/712

I'm going to close this, because I'd prefer a solution that doesn't introduce an extra setting. Plus this solution would require both Widgets Bundle and Page Builder to be active at the same time, which isn't always the case.

PR #544, opened 2018-06-07
Row Cell options: Prevent Yoast from resizing fields

AlexGStapleton:
This fixes an issue that Yoast introduced in the latest update where it'll resize the column size fields. This PR also prevents it from happening again because setting a width by default.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -1234,6 +1234,7 @@
 				input, select, button, strong, span {
 					display: inline;
 					margin: 1px 5px;
+					width: auto;
 
 					outline: none;
 					box-shadow: none;


Misplon:
@AlexGStapleton Would you also kindly log this on Yoast's repo. Really not cool targetting global admin elements like that. So easy to use a namespace.

AlexGStapleton:
Looks like this was [already fixed](https://github.com/Yoast/wordpress-seo/issues/9969) in the latest release so there's no need to rush this out. Should still keep this merged because any plugin can do the same and this PR just prevents it.

Misplon:
Sweet!

PR #551, opened 2018-06-22
Stopped duplicate revisions from being created in the wp_posts table.

chrisbarnett1:
This PR stops duplicate revisions from being created in the wp_posts table.

If you create a new page with the page builder an extra revision is saved for every update made. 

e.g. this page has been saved then updated but has a total of 4 revisions
duplicate-revisions

This PR stops the extra revision from been created by temporarily removing the save revision hook while wp_update_post() is called. 
duplicate-revisions-fix

See this on stackexchange for more details: https://wordpress.stackexchange.com/questions/3393/enable-disable-post-revisions-programmatically/3398

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -209,7 +209,9 @@ class SiteOrigin_Panels_Admin {
 					$post->post_content .= $post_css;
 					$post->post_content .= '</style>';
 				}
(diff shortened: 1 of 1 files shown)


braamgenis:
This issue seems to have been resolved by the change you made in #552 .

PR #552, opened 2018-06-22
Fixed bug with the layout IDs for revisions

chrisbarnett1:
Layout IDs should use the post ID rather than the revision ID to prevent WordPress from highlighting the ID as a change in the revision browser.

The "Content" section in the revision browser has the wrong layout IDs.
e.g.
wrong-id-in-revision-browser

The "Page Builder Content" section has things right imo. This uses the current post ID rather than revision ID. The Content section IMO should also use the post ID rather than revision ID. This PR fixes this so things are like this:
wrong-id-in-revision-browser-fix

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -193,10 +193,13 @@ class SiteOrigin_Panels_Admin {
 
 			if( siteorigin_panels_setting( 'copy-content' ) ) {
 				// Store a version of the HTML in post_content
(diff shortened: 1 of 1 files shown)


braamgenis:
Hi @chrisbarnett1 Thanks so much for your contribution and nicely done tracking down and fixing these issues.

PR #553, opened 2018-06-22
Stop duplicate panel_data from being added to wp_postmeta.

chrisbarnett1:
This PR fixes a bug where a single revision had multiple panel_data objects like this:
duplicate-panel-data

### DIFF
--- a/inc/revisions.php
+++ b/inc/revisions.php
@@ -34,10 +34,14 @@ class SiteOrigin_Panels_Revisions {
 
 		$parent_id = wp_is_post_revision( $post_id );
 		if ( $parent_id ) {
-			// If the panels data meta exists, copy it into the revision.
-			$panels_data = get_post_meta( $parent_id, 'panels_data', true );
-			if ( ! empty( $panels_data ) ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
It might be a better idea to just run `update_post_meta()` instead of `add_post_meta()`.

chrisbarnett1:
That might work. I thought the add if not set approach would be the safer option in this instance though.

chrisbarnett1:
I reviewed the data that was from the example of the bug I posted (i.e. the screenshot above). It seems the panels_data for meta_id 70 is different to what is in rows 71 and 72.

I believe when WordPress is saving an update, the revision is saved in the database before the original post. So an update post meta wouldn't work in this case.

The save_post hook in admin.php usually adds the panels_data to both the original post and the revision (which is why I initially tried removing this save_post hook in revisions.php). The process of restoring a revision (which in turn creates a new revision) seems to be slightly different. This is where I think this hook in revisions.php is still actually needed.

braamgenis:
Thanks for this @chrisbarnett1 👍

PR #565, opened 2018-07-25
Give post loop access to widget scope

SwiftThemes:
The templates used in the post loop have no idea about the width of the div they are used in or the index of post that they are displaying. We use manually include the template instead of autoloading it, thus giving it access to the widget instance. We also add width option to the widget instance, which is passed down to the template.

### DIFF
--- a/inc/widgets/post-loop-helper.php
+++ b/inc/widgets/post-loop-helper.php
@@ -6,24 +6,24 @@
  * Class SiteOrigin_Panels_Widgets_PostLoop_Helper
  */
 class SiteOrigin_Panels_Widgets_PostLoop_Helper extends SiteOrigin_Widget {
-	
+
 	/**
 	 * SiteOrigin_Panels_Widgets_PostLoop_Helper constructor.
 	 *
 	 * @param array $templates
 	 */
 	function __construct( $templates ) {
-		
+
 		$template_options = array();
-		if( ! empty( $templates ) ) {
-			foreach( $templates as $template ) {
-				$headers = get_file_data( locate_template( $template ), array(
+		if ( ! empty( $templates ) ) {
+			foreach ( $templates as $template ) {
+				$headers                       = get_file_data( locate_template( $template ), array(
 					'loop_name' => 'Loop Name',
 				) );
 				$template_options[ $template ] = esc_html( ! empty( $headers['loop_name'] ) ? $headers['loop_name'] : $template );
 			}
 		}
-		
+
 		parent::__construct(
 			'siteorigin-panels-postloop-helper',
 			__( 'Post Loop', 'siteorigin-panels' ),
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
I disagree with this change. I agree with should be passing the `$widget` instance to the template itself but using include to do that isn't the right move as that will prevent child themes from replacing the post loop template.

Misplon:
@SwiftThemes Hi Satish. Alex and I are busy looking at ways we can help, sorry it wasn't sooner!

AlexGStapleton:
Sorry for the delay.

I wasn't aware of this previously but, in 2.7.0 we introduced a method to access `$widget`. You can access `$widget` by using [SiteOrigin_Panels_Widgets_PostLoop::get_current_loop_instance()](https://github.com/siteorigin/siteorigin-panels/blob/2.7.3/inc/widgets/post-loop.php#L56).

There's no direct way to know what iteration the loop is on (whether its the 3rd of 9th post), so I would consider changing your PR to do that (which I agree, that would be handy) if you feel the above works. If it doesn't, consider submitting a PR with any adjustments to how you would like for it to work instead - we'll definitely consider them.

SwiftThemes:
Hi Alex, 
We are using include with locate_template. So child themes shouldn't be a problem.
We are looking for the file WordPress way, and including it manually so that it gets the scope outside. 
I haven't verified it with a child theme yet, will check and get back to you in a day.

Thanks
Satish Gandham

AlexGStapleton:
Hi Satish,

Ah, my mistake. Didn't notice you included `locate_template` include the include. Regardless, I'm not too sure if the inclusion should be made as it's possible to access $widget using  [SiteOrigin_Panels_Widgets_PostLoop::get_current_loop_instance()](https://github.com/siteorigin/siteorigin-panels/blob/2.7.3/inc/widgets/post-loop.php#L56).

Regarding the width setting. I'm not too sure if that setting should be inclued. I understand that there may be a specific use for it with your theme(s), but I don't see many other themes (if any) using that setting. I think [adding that field directly to your themes](https://siteorigin.com/docs/page-builder/hooks/custom-row-settings/) may be the better overall soloution as that way you'll get control over that setting completely.

SwiftThemes:
I wasn't aware of this before, SiteOrigin_Panels_Widgets_PostLoop::get_current_loop_instance(), I will use that.
Anyway, you can provide a filter to add additional fields to the widget?
If I have to make this change within the theme, I would probably have to clone the widget. Is there a better way?

AlexGStapleton:
The widget styles filter is the same as the [row filter](https://siteorigin.com/docs/page-builder/hooks/custom-row-settings/), but with widget instead of row - siteorigin_panels_widget_style_fields. Basically, as a row of thumb, all row instructions apply to widgets, just replace word row with the word widget and it'll adjust the widget instead of the row.

We've intentionally designed SiteOrigin Page Builder and the SiteOrigin Widgets Bundle so you don't need to clone widgets or anything like that. Hook into the relevant area and make the desired change.

I would recommend giving the [SiteOrigin Page Builder](https://siteorigin.com/docs/page-builder/) and [SiteOrigin Widgets Bundle](https://siteorigin.com/docs/widgets-bundle/) developer docs a quick read.

SwiftThemes:
post loop widget is a part of siteorigin-panels, and it doesn't seem to have this filter.
`add_filter('siteorigin_widgets_form_options_siteorigin-panels-postloop', 'helium_extend_siteorigin_panels_postloop', 10, 2);`

Can you please tell me the filter I should be using.

AlexGStapleton:
What specifically are you looking to modify? I ask as how you do that depends on what you want to do.
Regardless, a general filter you can use is `siteorigin_panels_widget_form`

```
function so_postloop_form_modify( $form, $widget, $instance ) {
	if ( $widget != 'SiteOrigin_Panels_Widgets_PostLoop' ) return $form;

	// Add custom code here

	return $form;
}
add_filter('siteorigin_panels_widget_form', 'test123', 10, 3);
```

SwiftThemes:
Thanks.
I'm looking to add additional fields.
Width to start with.

AlexGStapleton:
Ah. You _can_ use the above-mentioned filter but I wouldn't recommend it as it is a tad complicated. 

Your best bet would be to [create a basic widget](https://siteorigin.com/docs/widgets-bundle/getting-started/creating-a-widget/) and [add a Posts form field](https://siteorigin.com/docs/widgets-bundle/form-building/form-fields/#heading-posts) - this will require that the [SiteOrigin Widgets Bundle](https://wordpress.org/plugins/so-widgets-bundle/) is installed. You'll need ot [create a custom Post Loop](https://siteorigin.com/docs/widgets-bundle/form-building/post-selector/) (although there's nothing stopping you from using an existing template) but it'll allow for you to have more control over the loop in general.

Misplon:
Hi @SatishGandham Hope you're well. How are things going with your task? Have you made progress?

PR #575, opened 2018-08-26
Don't remove left/right border when Full Width Stretch Padding is enabled on row

AlexGStapleton:
This PR will prevent the Full Width Stretched Padding row layout from removing the right/left border on rows.

### DIFF
--- a/js/styling.js
+++ b/js/styling.js
@@ -44,8 +44,8 @@ jQuery( function ( $ ) {
 			}
 
 			$$.css( {
-				'border-left': 0,
-				'border-right': 0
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
[Slack](https://siteorigin.slack.com/archives/D0MGRNNTA/p1536175385000100)

I understand this isn't something we're looking to apply to the Full Width Stretched Row Layout, but as this PR is for the Full Width Stretched Padding Row Layout I don't see any issues with making this change .

AlexGStapleton:
All good mate. 👍

PR #576, opened 2018-08-27
Get CSS change to allow for only min-width to be set

AlexGStapleton:
Previously it wasn't possible to only specify a `min-width` with `get_css()` (which is used with `add_row_css()`). This PR will allow you to set only a `min-width` by passing a breakpoint of `0:768` (`768 `being the `min-width`).

### DIFF
--- a/inc/css-builder.php
+++ b/inc/css-builder.php
@@ -217,7 +217,7 @@ class SiteOrigin_Panels_Css_Builder {
 				continue;
 			}
 
-			if ( empty( $max_res ) && ! empty( $min_res ) ) {
+			if ( $max_res === '' && ! empty( $min_res ) ) {
 				$css_text .= '@media (min-width:' . intval( $min_res ) . 'px) {';
 			} elseif ( $max_res < 1920 ) {
 				$css_text .= '@media (max-width:' . intval( $max_res ) . 'px)';
--- a/inc/css-builder.php
+++ b/inc/css-builder.php
@@ -232,7 +232,7 @@ class SiteOrigin_Panels_Css_Builder {
 				$css_text .= implode( ' , ', $selector ) . ' { ' . $property . ' } ';
 			}
 
-			if ( ( $max_res == 0 && ! empty( $min_res ) ) ||  $max_res < 1920 ) {
+			if ( ( $max_res === 0 && ! empty( $min_res ) ) ||  $max_res < 1920 ) {
 				$css_text .= ' } ';
 			}
 		}
--- a/inc/css-builder.php
+++ b/inc/css-builder.php
@@ -217,7 +217,9 @@ class SiteOrigin_Panels_Css_Builder {
 				continue;
 			}
 
-			if ( $max_res < 1920 ) {
+			if ( $max_res == 0 && ! empty( $min_res ) ) {
+				$css_text .= '@media (min-width:' . intval( $min_res ) . 'px) {';
+			} elseif( $max_res < 1920 ) {
 				$css_text .= '@media (max-width:' . intval( $max_res ) . 'px)';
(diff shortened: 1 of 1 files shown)


inc/css-builder.php
```diff
@@ -217,7 +217,9 @@ public function get_css() {
 				continue;
 			}
 
-			if ( $max_res < 1920 ) {
+			if ( $max_res == 0 && ! empty( $min_res ) ) {
+				$css_text .= '@media (min-width:' . intval( $min_res ) . 'px) {';
+			} elseif( $max_res < 1920 ) {
```
Misplon: Perhaps can add a space: `} elseif (`

inc/css-builder.php
```diff
@@ -217,7 +217,9 @@ public function get_css() {
...
 				$css_text .= '@media (max-width:' . intval( $max_res ) . 'px)';
 				if( ! empty( $min_res ) ) {
```
Misplon: And one more `if ( ! empty( $min_res ) ) {`

AlexGStapleton: Can you please clarify what line you're suggesting that gets added to? I ask as it's checked on 220 and 223 so I'm not too sure what you're suggesting it needs to be added. Sorry about this.

Misplon: Line 224.

inc/css-builder.php
```diff
@@ -217,9 +217,11 @@ public function get_css() {
 				continue;
 			}
 
-			if ( $max_res < 1920 ) {
+			if ( $max_res == 0 && ! empty( $min_res ) ) {
```
braamgenis: It seems like this is using `0` to say there isn't a `max-width`, which could be confusing, so rather use an empty string. E.g. instead of `0:768` you'll use `:768`.
Then this condition would be `if ( $max_res === '' && $min_res > 0 ) {` or use `$min_res >= 0` if you want to include `0` here

AlexGStapleton:
I couldn't find any reasoning for not allowing just a `min-width` outside of just not having a use for it.

inc/css-builder.php
```diff
@@ -217,9 +217,11 @@ public function get_css() {
 				continue;
 			}
 
-			if ( $max_res < 1920 ) {
+			if ( empty( $max_res ) && ! empty( $min_res ) ) {
```
braamgenis: We need to specifically check for `$max_res === ''` because `0` is also considered an empty value.

AlexGStapleton: That was actually intentional on my part (as I think `0` instead of nothing is neater), but if you would rather not allow `0` at all that's fine - I'll make the change now.

braamgenis: `0` is a valid number and should be allowed as a value, but it doesn't mean the same as 'not set'.

AlexGStapleton: Just to confirm, no further action is required, right? I ask as the last commit checks for`$max_res === ''`

braamgenis: Ideally, this would be `if ( $max_res === '' && $min_res > 0 )` as that's more precisely what we're trying to check here, but I think this will work as is.

inc/css-builder.php
```diff
@@ -230,7 +232,7 @@ public function get_css() {
 				$css_text .= implode( ' , ', $selector ) . ' { ' . $property . ' } ';
 			}
 
-			if ( $max_res < 1920 ) {
+			if ( ( $max_res === 0 && ! empty( $min_res ) ) ||  $max_res < 1920 ) {
```
braamgenis: The first part of this should be using the same logic as above, shouldn't it?

braamgenis:
Cool, that should do it.

PR #579, opened 2018-09-23
Sidebar Emulator: Account for widgets that don't use ids and instead rely on widget class

AlexGStapleton:
The [Jetpack Milestone](https://jetpack.com/support/extra-sidebar-widgets/milestone-widget/) widget [uses a classnames](https://github.com/Automattic/jetpack/blob/6.5/modules/widgets/milestone/milestone.php#L37) instead of an id and the sidebar emulator doesn't support classes. This PR fixes this and allows for the Jetpack Milestone widget and any other widget that relies on Class Names to work.

Note: The priority for the sidebar emulation code was changed due to Jetpack running it's is_widget_active() check before we set up the sidebar emulation. I couldn't find an explanation as to why the priority was changed so I reverted the priority to default - this may have unintended consequences but I wasn't able to find any during testing.

### DIFF
--- a/inc/sidebars-emulator.php
+++ b/inc/sidebars-emulator.php
@@ -6,7 +6,7 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 
 	function __construct() {
 		$this->all_posts_widgets = array();
-		add_action( 'widgets_init', array( $this, 'register_widgets' ), 99 );
+		add_action( 'widgets_init', array( $this, 'register_widgets' ) );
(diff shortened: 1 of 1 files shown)


braamgenis:
Please provide some more info on the issue that this PR is solving? In what way does the JP Milestone Widget not work when using the Sidebar Emulator? Some steps to reproduce, expected and actual behavior would also help me to understand the issue, test and confirm before reviewing the code. Thanks!

AlexGStapleton:
Please disregard this PR as it's not a proper solution. I'm going to log this issue with more information shortly.

PR #583, opened 2018-10-02
Auto text widget excerpt

### LINKED ISSUE #581: Automatic excerpts from first SiteOrigin Editor or WP Text widget
From siteorigin/so-widgets-bundle#749:

Would be a nice addition to automatically create excerpts for Page Builder pages, by extracting the content of the first texty type widget. We'll start with SiteOrigin Editor or WP Text widgets.

We should limit the search for a text widget to the first row or two, to prevent an excerpt being created from content which is far down in the page.

braamgenis:
To resolve #581 . This just copies the default excerpt functionality, but applies it to the first text type widget found in the first two PB rows.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -355,7 +355,7 @@ class SiteOrigin_Panels {
 		if ( $panels_data && ! empty( $panels_data['widgets'] ) ) {
 			$raw_excerpt = '';
 			$excerpt_length = apply_filters( 'excerpt_length', 55 );
-			foreach ( $panels_data['widgets'] as $i => $widget ) {
+			foreach ( $panels_data['widgets'] as $widget ) {
 				$panels_info = $widget['panels_info'];
 				if ( $panels_info['grid'] > 1 ) {
 					// Limiting search for a text type widget to the first two PB rows to avoid having excerpt content
@@ -363,7 +363,7 @@ class SiteOrigin_Panels {
 					break;
 				}
 				if ( $panels_info['class'] == 'SiteOrigin_Widget_Editor_Widget' || $panels_info['class'] == 'WP_Widget_Text' ) {
-					$raw_excerpt .= ( $i > 0 ? ' ' : '' ) . $widget['text'];
+					$raw_excerpt .= ' ' . $widget['text'];
 					// This is all effectively default behavior for excerpts, copied from the `wp_trim_excerpt` function.
 					// We're just applying it to text type widgets content in the first two rows.
 					$text = strip_shortcodes( $raw_excerpt );
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -65,7 +65,7 @@ class SiteOrigin_Panels {
 		add_filter( 'wp_enqueue_scripts', array( $this, 'generate_post_css' ) );
 		
 		// Remove the default excerpt function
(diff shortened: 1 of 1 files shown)


siteorigin-panels.php
```diff
@@ -343,20 +335,83 @@ public function generate_post_content( $content ) {
...
+				if ( $panels_info['grid'] > 1 ) {
+					// Limiting search for a text type widget to the first two PB rows to avoid having excerpt content
+					// that's very far down in a post.
+					break;
+				}
+				if ( $panels_info['class'] == 'SiteOrigin_Widget_Editor_Widget' || $panels_info['class'] == 'WP_Widget_Text' ) {
+					$text = $widget['text'];
+					// This is all effectively default behavior for excerpts, copied from the `wp_trim_excerpt` function.
+					// We're just applying it to the first text type widget's content.
+					$raw_excerpt = $text;
```
gregpriday: Sometime I just thought of here. What if the user has spread their opening content across a few text widgets. This block should just keep adding to the `$raw_excerpt` variable until its length is >= `$excerpt_length`. Once that happens, it should break out of the `foreach` loop.

All the rest of the processing should be applied to `$raw_excerpt` after the `foreach` loop.

braamgenis: Hmmm, yeah good point. Users could use various different configurations of text widgets, so perhaps we need to allow them to specify exactly which widgets they might want as excerpts with a checkbox in Widget Settings?

Alternatively, perhaps we should give them a way to switch the auto-excerpt functionality? Just slightly concerned that we might end up using things as excerpts that don't make sense. E.g. what if a user is using the Editor Widget just for headings, or images, or some other HTML?

gregpriday: I think just relying on any text-like widgets is good enough. If anything is really weird with the excerpt, then the solution is to manually write one.

The auto-excerpt should be 100% automatic, so I'd be against adding any settings/checkboxes. What we're trying to do here is make the automaticness slightly more intelligent.

braamgenis: Ok, cool. Then I'd just need to change the behavior slightly. Need to run our filter earlier than the default filter and then if there's an existing manual excerpt, it will just return that.

siteorigin-panels.php
```diff
@@ -353,33 +353,49 @@ public function generate_post_excerpt( $text ) {
...
+			foreach ( $panels_data['widgets'] as $i => $widget ) {
 				$panels_info = $widget['panels_info'];
 				if ( $panels_info['grid'] > 1 ) {
 					// Limiting search for a text type widget to the first two PB rows to avoid having excerpt content
 					// that's very far down in a post.
 					break;
 				}
 				if ( $panels_info['class'] == 'SiteOrigin_Widget_Editor_Widget' || $panels_info['class'] == 'WP_Widget_Text' ) {
-					$text = $widget['text'];
+					$raw_excerpt .= ( $i > 0 ? ' ' : '' ) . $widget['text'];
```
gregpriday: Very minor, but if the first widget is a non text widget, then this ternary will add a space at the start of the next text widget's text. You could drop the ternary, just make it always add a space and call `trim` just before `wp_trim_words`. I haven't checked properly, but `wp_trim_words` probably does its own trimming, so that might not even be necessary.

braamgenis: Hah, nice catch. Exactly what I was trying to avoid. :finnadie:

PR #585, opened 2018-10-06
Fix sidebar emulator for Jetpack widgets.

braamgenis:
Need to add widgets' `option_` filters earlier and generate and store ids for the emulated sidebar in `save_post` to allow for `is_active_widget` to be checked by widgets when they are registered.

### DIFF
--- a/inc/sidebars-emulator.php
+++ b/inc/sidebars-emulator.php
@@ -92,12 +92,10 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 	 *
 	 * @param int|string $post_id
 	 * @param array $panels_data
-	 * @param int $start This keeps track of recursive depth
 	 *
 	 * @return array A list of widget option names from the post and its Layout Builder widgets.
 	 */
-	private function get_widget_option_names( $post_id, $panels_data, $start = 1 ) {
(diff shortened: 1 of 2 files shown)


inc/sidebars-emulator.php
```diff
@@ -115,20 +113,50 @@ private function get_widget_option_names( $post_id, $panels_data, $start = 1 ) {
...
+	 * do this separately for widgets that call `is_active_widget` in their constructors, e.g. some of Jetpack's widgets
+	 * like Twitter Timeline, Milestone etc.
+	 *
+	 * @param $widgets array The widgets in the layout from $panels_data
+	 * @param $post_id int The post id.
+	 * @param int $start This keeps track of recursive depth.
+	 *
+	 * @return array The widgets array containing updated widgets.
+	 */
+	public function set_widget_ids( $widgets, $post_id, $start = 1 ) {
```
gregpriday: This function name is a touch confusing. I don't know if it's just me, but when I see a function name with `set_` I expect it to set something in the class. I'd expect this function to be called `add_widget_ids` or something.

Could just be a personal preference, so this doesn't really need to change.

PR #587, opened 2018-10-09
Max Mega Menu: Prevent layout builder duplication

### LINKED ISSUE #573: [2.5] Max Mega Menu: Layout Builder opens multiple times
Since 2.5, [Max Mega Menu](https://en-au.wordpress.org/plugins/megamenu/) (200k) has issues with the layout builder widget. Opening the builder will result in three SiteOrigin dialog overlays. [Here's a video of this issue.](https://drive.google.com/a/siteorigin.com/uc?id=1pH9P3g7lWuHgcpe7TfH4o4qLqtdK9xAQ)

According to the markup, there are three instances of the `.so-panels-dialog-wrapper`.

AlexGStapleton:
Resolve #573 

Max Mega Menu loads the full widget form which includes the `soPanelsSetupBuilderWidget();` setup code in inc/widgets/layout.php.

Max Mega Menu always triggers `widget-added` upon succesful widget load. This is good but as we run `soPanelsSetupBuilderWidget();` when that event is triggered it results in mutiple instances of SiteOrigin Page Builder being setup

[](https://github.com/siteorigin/siteorigin-panels/blob/2.8.2/js/siteorigin-panels/main.js#L145) widget-added on load.

### DIFF
--- a/inc/widgets/layout.php
+++ b/inc/widgets/layout.php
@@ -116,11 +116,13 @@ class SiteOrigin_Panels_Widgets_Layout extends WP_Widget {
 			<input type="hidden" value="<?php echo esc_attr( $instance['builder_id'] ) ?>" name="<?php echo $this->get_field_name('builder_id') ?>" />
 		</div>
 		<script type="text/javascript">
-			if(
+			if (
 				typeof jQuery.fn.soPanelsSetupBuilderWidget != 'undefined' &&
-				( ! jQuery('body').hasClass('wp-customizer') || jQuery( "#siteorigin-page-builder-widget-<?php echo esc_attr( $form_id ) ?>").closest( '.panel-dialog' ).length )
+				( 
+					( ! jQuery( 'body' ).hasClass( 'wp-customizer' ) && ! jQuery( 'body' ).hasClass( 'megamenu_enabled' ) )
+					|| jQuery( "#siteorigin-page-builder-widget-<?php echo esc_attr( $form_id ) ?>").closest( '.panel-dialog' ).length )
 			) {
-				jQuery( "#siteorigin-page-builder-widget-<?php echo esc_attr( $form_id ) ?>").soPanelsSetupBuilderWidget();
+				jQuery( "#siteorigin-page-builder-widget-<?php echo esc_attr( $form_id ) ?>" ).soPanelsSetupBuilderWidget();
 			}
 		</script>
 		<?php


AlexGStapleton:
I'm personally not too worried about that situation happening as this is the first time this sort of issue has come up (it only happened after they changed how they load widget forms).

I understand your reasoning, but I think it would be a better idea to remove `soPanelsSetupBuilderWidget`  from layouts.php and instead trigger `widgets-added`.

braamgenis:
> I'm personally not too worried about that situation happening as this is the first time this sort of issue has come up (it only happened after they changed how they load widget forms).

Is it the first time it's come up? Why do we also have a check in place there for the customizer? Regardless, a good solution should fix the reported issue and try to prevent possible future issues, within reason. Just adding a check for `megamenu_enabled` will only solve the problem for the reported issue. If another 3rd party plugin has a widget area which triggers `widget-added`, the problem will occur again.

> I understand your reasoning, but I think it would be a better idea to remove `soPanelsSetupBuilderWidget` from layouts.php and instead trigger `widgets-added` - have all instances of SiteOrigin Page Builder be triggered by the same event.

That's a much better solution than just adding more and more checks here for plugins. However, the potential for triggering multiple setups of Page Builder widgets remains. We can't control always control when `widget-added` is triggered, and we can't prevent `soPanelsSetupBuilderWidget` from being called directly multiple times for a particular widget. A more robust solution is to have a way of checking and resetting the _initialized_ state of Layout Builder widgets.

AlexGStapleton:
Gonna leave this one up to you.

braamgenis:
A quick update on the conversation: I looked into the option of triggering the `widget-added` event when we add widgets in PB. While it is definitely the ideal solution, it's going to take quite some effort to implement.

The main challenge is that when WP triggers `widget-added` it provides the jQuery widget element which contains some contextual information about the widget. PB doesn't render or model widgets in exactly the same way as WP, so to trigger `widget-added` correctly we would need to wrap the widget and make sure it has all the same information as when WP triggers the `widget-added` event.

TL;DR: Retrofitting the `widget-added` event in PB would be hard and it should be a last resort solution, even though it's likely the most "correct" solution. :finnadie:

AlexGStapleton:
That's a shame. So the nitially suggested solution you suggested would be the best course of action?

braamgenis:
Yeah, it should hopefully prevent further similar issues.

PR #589, opened 2018-10-15
Post Loop: Fix post title and inclusion issue

AlexGStapleton:
Follow up to #559.

This PR will allow for the custom post loop to load in all situations and allow for its file name to appear in the template drop down.

### DIFF
--- a/inc/widgets/post-loop.php
+++ b/inc/widgets/post-loop.php
@@ -180,17 +180,16 @@ class SiteOrigin_Panels_Widgets_PostLoop extends WP_Widget {
 			echo $args['before_title'] . $instance['title'] . $args['after_title'];
 		}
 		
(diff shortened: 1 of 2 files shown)


Misplon:
@braamgenis This is to assist one of our developer license users loading Post Loop templates in a plugin. Thanks mate.

inc/widgets/post-loop.php
```diff
@@ -189,9 +189,10 @@ function widget( $args, $instance ) {
 				the_post();
 				locate_template($instance['template'], true, false);
 			}
-		}
-		else {
-			locate_template($instance['template'], true, false);
+		} else {
+			if( file_exists( $instance['template'] ) ) {
```
braamgenis: Could just be `elseif`

PR #592, opened 2018-10-16
Release 2.9.1

braamgenis:
* Fix auto-excerpt output.
* Layout builder: Fix 'undefined index' when saving before having added any widgets.
* Layout builder: Prevent initializing multiple instances of widget dialog.
* Post Loop: Allow templates from custom locations and display them in dropdown field.
* Prevent notices when style field is using 'label' instead of 'name' e.g. for checkbox field.

### DIFF
--- a/inc/sidebars-emulator.php
+++ b/inc/sidebars-emulator.php
@@ -6,7 +6,7 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 
 	function __construct() {
 		$this->all_posts_widgets = array();
-		add_action( 'widgets_init', array( $this, 'register_widgets' ), 99 );
+		add_action( 'widgets_init', array( $this, 'register_widgets' ) );
 		add_filter( 'sidebars_widgets', array( $this, 'add_widgets_to_sidebars' ) );
 	}
 
@@ -79,7 +79,7 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 		}
 
 		$panels_data = get_post_meta( $post_id, 'panels_data', true );
-		$widget_option_names = $this->get_widget_option_names( $post_id, $panels_data, 1 );
+		$widget_option_names = $this->get_widget_option_names( $post_id, $panels_data );
 		$widget_option_names = array_unique( $widget_option_names );
 
 		foreach ( $widget_option_names as $widget_option_name ) {
@@ -92,12 +92,10 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 	 *
 	 * @param int|string $post_id
 	 * @param array $panels_data
-	 * @param int $start This keeps track of recursive depth
 	 *
 	 * @return array A list of widget option names from the post and its Layout Builder widgets.
 	 */
-	private function get_widget_option_names( $post_id, $panels_data, $start = 1 ) {
-		global $wp_widget_factory;
+	private function get_widget_option_names( $post_id, $panels_data ) {
 		if( empty( $panels_data ) || empty( $panels_data[ 'widgets' ] ) ) {
 			return array();
 		}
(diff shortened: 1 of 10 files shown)


inc/widgets/post-loop-helper.php
```diff
@@ -17,10 +17,10 @@ function __construct( $templates ) {
 		$template_options = array();
 		if( ! empty( $templates ) ) {
 			foreach( $templates as $template ) {
-				$headers = get_file_data( locate_template( $template ), array(
```
gregpriday: Are we using `locate_template` somewhere else? We're using this to allow child themes to override templates, so it's fairly important.

inc/widgets/post-loop.php
```diff
@@ -180,18 +180,18 @@ function widget( $args, $instance ) {
...
 			while( have_posts() ) {
 				the_post();
 				locate_template($instance['template'], true, false);
 			}
-		}
-		else {
-			locate_template($instance['template'], true, false);
+		} elseif ( file_exists( $instance['template'] ) ) {
+				load_template( $instance['template'], false );
+			}
```
gregpriday: I assume there isn't a syntax error happening here, but this indenting doesn't make sense. Looks like this closing `}` shouldn't be here. I can't quite figure out what's going on.

inc/widgets/post-loop.php
```diff
@@ -252,10 +252,11 @@ function form( $instance ) {
 					<?php foreach($templates as $template) : ?>
 						<option value="<?php echo esc_attr($template) ?>" <?php selected($instance['template'], $template) ?>>
 							<?php
-							$headers = get_file_data( locate_template($template), array(
```
gregpriday: Here again, are we using a different mechanism somewhere to allow child themes to override post loop templates?

inc/sidebars-emulator.php
```diff
@@ -145,8 +145,10 @@ public function generate_sidebar_widget_ids( $widgets, $post_id, $start = 1 ) {
 			
 			
 			if( $widget_instance['panels_info']['class'] === 'SiteOrigin_Panels_Widgets_Layout' ) {
+				if ( ! empty( $widget_instance['panels_data']['widgets'] ) ) {
 				// Recursively set widget ids in layout widgets.
```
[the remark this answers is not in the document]
braamgenis: It is. 👍

inc/styles-admin.php
```diff
@@ -142,7 +142,9 @@ function render_styles_fields( $section, $before = '', $after = '', $current = a
 						if ( $field['group'] == $group_id ) {
 							?>
 							<div class="style-field-wrapper">
+								<?php if ( ! empty( $field['name'] ) ) : ?>
 								<label><?php echo $field['name'] ?></label>
```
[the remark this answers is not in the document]
braamgenis: Indenting in mixed PHP/HTML/templatey type files has always been a bit weird for me. It seems either the PHP or the HTML indentation is going to be off, so it's a question of which do you prefer?

inc/widgets/post-loop.php
braamgenis: Ah yeah, it will cause a syntax error. I missed this before merging the PR, then fixed it in the release branch locally, but didn't push the commit.

inc/widgets/post-loop-helper.php
braamgenis: The `SiteOrigin_Panels_Widgets_PostLoop_Helper` class is being instantiated with an array of the templates which are retrieved by the `get_loop_templates` function, which looks like it should load both parent and child theme templates:
https://github.com/siteorigin/siteorigin-panels/blob/900a83a19b6904104a864a2a0394cbb065322f93/inc/widgets/post-loop.php#L365-L366

gregpriday: Looking at `locate_template` [here](https://core.trac.wordpress.org/browser/tags/4.9.8/src/wp-includes/template.php#L629), it looks like it's loading from `ABSPATH . WPINC . '/theme-compat/'` too. So what we have is mostly compatible, but not exactly.

It's minor though and shouldn't cause any issues. As long as we've checked this with themes and child themes, we should be good.

PR #623, opened 2018-12-17
Add Radio Styles field

### LINKED ISSUE #620: customisation: radio buttons in Row Styles
Hi guys,
Currently it is only possible to use radio buttons in widgets (using the W. Bundle), but not row styles. 
This doesn't work out of the box for example:

```
<?php function custom_radio_fields($fields) {

	$fields['radio_option'] = array(
		'name'        => esc_html__( 'Name', 'textdomain' ),
		'type'        => 'radio',
		'group'       => 'design',
		'priority'    => 10,
		'options' => array(
			'1' => esc_html__( '1', 'textdomain' ),
			'2' => esc_html__( '2', 'textdomain' ),
			'3' => esc_html__( '3', 'textdomain' ),
		),
		'default'  => '1'
	);
	return $fields;
}
add_filter('siteorigin_panels_row_style_fields', 'custom_radio_fields', 10, 2);

```
Is there any filter that I can use to add radio button capability into the Page builder Row styles section?

AlexGStapleton:
Resolves #620.

### DIFF
--- a/inc/styles-admin.php
+++ b/inc/styles-admin.php
@@ -286,7 +286,7 @@ class SiteOrigin_Panels_Styles_Admin {
 				foreach ( $field['options'] as $k => $v ) : ?>
 					<label for="<?php echo esc_attr( $field_name . '-' . $k ) ?>">
 						<input type="radio" name="<?php echo esc_attr( $field_name ) ?>"
-					       id="<?php echo esc_attr( $field_name . '-' . $k ) ?>" class="siteorigin-widget-input"
+					       id="<?php echo esc_attr( $field_name . '-' . $k ) ?>"
 					       value="<?php echo esc_attr( $k ) ?>" <?php checked( $k, $current ) ?>> <?php echo esc_html( $v ) ?>
 					</label>
 					<?php
--- a/inc/styles-admin.php
+++ b/inc/styles-admin.php
@@ -283,10 +283,12 @@ class SiteOrigin_Panels_Styles_Admin {
 				break;
 
 			case 'radio' :
-				foreach ( $field['options'] as $k => $v ) : ?>
-					<label for="<?php echo esc_attr( $field_name . '-' . $k ) ?>">
-						<input type="radio" name="<?php echo esc_attr( $field_name ) ?>"
-					       id="<?php echo esc_attr( $field_name . '-' . $k ) ?>"
+				$radio_id = $field_name . '-' . uniqid();
+				foreach ( $field['options'] as $k => $v ) :
+					?>
+					<label for="<?php echo esc_attr( $radio_id . '-' . $k ) ?>">
+						<input type="radio" name="<?php echo esc_attr( $radio_id ) ?>"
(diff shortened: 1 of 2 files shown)


inc/styles-admin.php
```diff
@@ -282,6 +282,17 @@ class="image-fallback widefat" />
 				<?php
 				break;
 
+			case 'radio' :
+				foreach ( $field['options'] as $k => $v ) : ?>
+					<label for="<?php echo esc_attr( $field_name . '-' . $k ) ?>">
```
braamgenis: I briefly looked into this recently and it quickly became more effort than it's worth, seeing as a `radio` input could just be replaced with a `select` input. The problem with radio inputs is that they need to have unique names and this will generate the same names for each Page Builder row, and then the browser thinks they're all part of the same radio input group. Check what happens when you attempt to use this and click the radio input labels for different Page Builder rows.

You could try generating a unique id to append/prepend to each of these, but again, IMO it's not worth spending a lot of time on this as `select` inputs are a perfectly good substitute.

andrejarh: Hi, 
Radio buttons can't always be replaced by a select input.
Let's take this usecase for example:
If I want to give user a simple palette of theme defined colors, I can use and style the radio buttons:
For such use case radio buttons are much more appropriate then select, as all the possibilities are shown to the user.
Here is a simple demo:
https://jmp.sh/VKvDxd2

I could add the colors to the IRIS palete color picker, but if theme color changes later, then the user would need to chage it everywhere all over again. Now simply the background class is added to the cell and the color will always match the theme color.

For the technical part of unique namings, you can consider something like this:

```
<?php
case 'radio' :
	$radio_unique_id = uniqid() . '-';
	foreach ( $field['options'] as $k => $v ) : ?>
		<label for="<?php echo esc_attr( $field_name . '-' . $k ) ?>">
			<input type="radio" name="<?php echo esc_attr($radio_unique_id) . esc_attr( $field_name ) ?>"
				id="<?php echo esc_attr( $field_name . '-' . $k ) ?>" class="siteorigin-widget-input"
				value="<?php echo esc_attr( $k ) ?>" <?php checked( $k, $current ) ?>> <?php echo esc_html( $v ) ?>
		</label>
		<?php
	endforeach;
	break;
```

AlexGStapleton: How does that look? Not too sure how these fields would get reliably sanitized though.

braamgenis:
Looks ok. Is it working as expected? The case that was problematic previously is when you have multiple radio fields and try to use the same radio fields in different rows on a page. Also need to make sure clicking on labels works as expected.

braamgenis:
Oh and as for the sanitization, you'll need to add another `case` for the `radio` type in the `sanitize_style_fields` function [here](https://github.com/siteorigin/siteorigin-panels/blob/2.9.7/inc/styles-admin.php#L419).

AlexGStapleton:
It's working as expected for me - fields are unique and aren't affected by selections in other rows and the labels are working for me.

PR #634, opened 2019-01-30
Prevent running widget logic while in wp_head

lucasrcosta:
When Yoast SEO is installed but no description is set on the page where Page Builder is configured, the widget logic runs unnecessarily, when the Yoast is processing the metadata.

### DIFF
--- a/inc/widgets/post-loop.php
+++ b/inc/widgets/post-loop.php
@@ -81,6 +81,7 @@ class SiteOrigin_Panels_Widgets_PostLoop extends WP_Widget {
 	 */
 	function widget( $args, $instance ) {
 		if( empty( $instance['template'] ) ) return;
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
Hi Lucas,

Thanks for your interest and contribution. Does this issue only occur when the Post Loop widget is present? I ask as the provided PR only affects the post loop widget and not any other widgets.

AlexGStapleton:
Thanks. I've looked into this and Yoast [uses get_the_excerpt](https://github.com/Yoast/wordpress-seo/blob/trunk/frontend/class-opengraph.php#L605) for generating the meta description which is why this is occurring (that uses the_content so it triggers SiteOrigin Page Builder). This happens for all widgets so this fix, while it's technically valid, doesn't prevent this happen globally. A global method of avoiding this is more ideal.

PR #637, opened 2019-02-04
Fix notice when using WP 4.9.9

AlexGStapleton:
This change prevents the below notice from occurring whenever `SiteOrigin_Panels_Admin::is_admin()` is used when using 4.9.9.

`019 15:45:24 UTC] PHP Notice:  Undefined property: WP_Screen::$is_block_editor in C:\wamp64\www\siteorigin\wp-4-9-9\wp-content\plugins\siteorigin-panels\inc\admin.php on line 118
`
Based on the previous checks, there doesn't appear to be a reliable manner to check for the Gutenberg plugin pre-WordPress 5.0. There doesn't appear to be any issues by letting it pass.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -115,7 +115,7 @@ class SiteOrigin_Panels_Admin {
 		$screen         = get_current_screen();
 		$is_panels_page = ( $screen->base == 'post' && in_array( $screen->id, siteorigin_panels_setting( 'post-types' ) ) ) ||
 						  in_array( $screen->base, array( 'appearance_page_so_panels_home_page', 'widgets', 'customize' ) ) ||
-						  $screen->is_block_editor;
+						  ( ! method_exists( $screen, 'is_block_editor' ) || $screen->is_block_editor );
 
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
[Reported here](https://siteorigin.com/thread/error-message-undefined-property/).

AlexGStapleton:
How does that look?

inc/admin.php
```diff
@@ -115,11 +115,29 @@ static function is_admin() {
...
+	 * Check if the current page is Gutenberg or the Block Ediotr
+	 *
+	 * @return bool
+	 */
+	static function is_block_editor() {
+		// This is for the Gutenberg plugin.
+		$is_gutenberg_page = function_exists( 'is_gutenberg_page' ) && is_gutenberg_page();
+		// This is for WP 5 with the integrated block editor.
+		$is_block_editor = false;
+		$current_screen = get_current_screen();
```
braamgenis: Seeing as this is now in a separate function, it has no knowledge of context and this function might not exist when called. Let's add a `function_exists` check here too to be safe.

PR #642, opened 2019-02-18
Pass SiteOrigin Image Grid, Masonry, and Button widget data to Yoast

AlexGStapleton:
This PR will provide Yoast with all of the images used in the Image Grid and Masonry widgets.

This PR will also provide Yoast with the URL used with Yoast. As with the images, however, it doesn't resolve the link so it won't be enough to resolve the Yoast internal flag unless the user doesn't set the link using the Select Content option.

### DIFF
--- a/js/yoast-compat.js
+++ b/js/yoast-compat.js
@@ -117,6 +117,12 @@ jQuery(function($){
 								}
 								newHTML = newHTML.prop( 'outerHTML' );
 								break;
+							case 'SiteOrigin_Widget_Button_Widget':
+								var hrefSeparator = widgetInstance.url.includes( '://' ) === true ? '' : '#';
+								newHTML = $( '<a>' + widgetInstance.text + '</a>' ).attr({
+									'href': hrefSeparator + widgetInstance.url,
+								}).prop('outerHTML');
+								break;
 						}
 					}
 
--- a/js/yoast-compat.js
+++ b/js/yoast-compat.js
@@ -84,6 +84,24 @@ jQuery(function($){
 								}).prop('outerHTML');
 								break;
 
+							case 'SiteOrigin_Widgets_ImageGrid_Widget':
+							case 'SiteOrigin_Widget_Simple_Masonry_Widget':
+								newHTML = $( '<div/>' );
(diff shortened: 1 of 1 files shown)


js/yoast-compat.js
```diff
@@ -99,6 +117,12 @@ jQuery(function($){
 								}
 								newHTML = newHTML.prop( 'outerHTML' );
 								break;
+							case 'SiteOrigin_Widget_Button_Widget':
+								var hrefSeparator = widgetInstance.url.includes( '://' ) === true ? '' : '#';
```
gregpriday: Is `=== true` necessary? This function returns either `true` or `false`, so it could just be `widgetInstance.url.includes( '://' ) ? '' : '#'`

js/yoast-compat.js
```diff
@@ -84,6 +84,24 @@ jQuery(function($){
...
 								break;
 
+							case 'SiteOrigin_Widgets_ImageGrid_Widget':
+							case 'SiteOrigin_Widget_Simple_Masonry_Widget':
+								newHTML = $( '<div/>' );
+								var contentItems = widgetClass === 'SiteOrigin_Widgets_ImageGrid_Widget' ? widgetInstance.images : widgetInstance.items;
+								for( var i = 0; i < contentItems.length; i++ ) {
+									var item = contentItems[ i ];
+									var itemHTML = $('<img/>').attr({
+									'src': '#' + item.image,
```
gregpriday: These lines in the associative array should be indented one level.

js/yoast-compat.js
AlexGStapleton: I was basing it on [this](https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#equality). If you would like me to change it, let me know and I'll change it - I personally don't mind.

gregpriday: Yeah, as a general rule it's better to use `===`, but in this case, even using `==` would be incorrect. Functions that return booleans, don't need to be compared to a boolean except in a few special cases. Like if `includes` returned `false` if the string didn't exist and an integer position if it did exist, where `0` was a valid position. Then `!== false` would be okay.

Unnecessary code makes things harder to read.

AlexGStapleton: That's understandable and makes sense. I'll make the change.

PR #648, opened 2019-02-26
Release 2.10.2

braamgenis:
* Don't remove left/right border when Full Width Stretch Padding is enabled on row.
* Display widget count for inline-save.
* Live editor: Press escape to close.
* Live editor: Give the user an option to either close or close and save.
* Added widget class to widgets in builder interface.
* Dialog crumbtrail fix.
* Only close topmost Page Builder window when escape key is pressed.
* Layout Block: Retrieve sanitized panels data from server as changes are made.

### DIFF
--- a/tpl/js-templates.php
+++ b/tpl/js-templates.php
@@ -133,7 +133,7 @@ $layouts = apply_filters( 'siteorigin_panels_prebuilt_layouts', array() );
 </script>
 
 <script type="text/template" id="siteorigin-panels-builder-widget">
-	<div class="so-widget ui-draggable">
+	<div class="so-widget ui-draggable" data-widget-class="{{%- widget_class %}}">
 		<div class="so-widget-wrapper">
 			<div class="title">
 				<h4>{{%= title %}}</h4>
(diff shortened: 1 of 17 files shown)


tpl/js-templates.php
```diff
@@ -551,7 +551,8 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ) ?>{{% if( rowColorLabel =
 		</div>
 
 		<div class="so-sidebar-tools">
-			<button class="live-editor-close button-primary"><?php esc_html_e('Done', 'siteorigin-panels') ?></button>
+			<button class="live-editor-save button-primary"><?php esc_html_e('Save', 'siteorigin-panels') ?></button>
```
braamgenis: @gregpriday Perhaps this buttons label should be _Save Draft_ for new posts and _Update Post_ for existing posts?

gregpriday: I'll do that fix soon.

gregpriday:
@braamgenis we also need to include https://github.com/siteorigin/siteorigin-panels/commit/17518e4ab369e74f67422349272d86280da712fc - small, but have to keep our tested tag at 5.1

PR #649, opened 2019-03-04
Feature/welcome page

Misplon:
@gregpriday We could use the above the fold space better on this page if we converted the content from rows into columns. And perhaps with a thumbnail next to each feature. If you're ok with that change, let me know how you want to go about it. Are there styles we can use from the core or can I add the columns to our admin stylesheet? Thanks.


gregpriday:
I'm keen to just go with a simple page to start and then expand it over time. The first thing I noticed is that the welcome page text needs to be run through translation functions.

Misplon:
Indeed it does! Epic, thanks for the feedback. Will hit this ASAP.

PR #652, opened 2019-03-05
Cache the result of url_to_postid()

WPprodigy:
The `url_to_postid()` function includes a fairly expensive, uncached, WP query. This can happen on most to all pages with this plugin, which hurts performance significantly.

I spent a little bit of time to see if we could avoid this entirely with a better solution for getting the `$post_id` in this function, one that doesn't involve a query based on the current request URL. I didn't come across any simple changes, but open to ideas on that front.

As a fix for the performance problems here, I've implemented some caching so `url_to_postid()` isn't repeatedly called. This is similar to the helper function we use at WordPress.com VIP: https://github.com/Automattic/vip-go-mu-plugins/blob/da7fcd24b773e1957687c6eb56f0b56ddd402d9e/vip-helpers/vip-caching.php#L252-L276

### DIFF
--- a/inc/sidebars-emulator.php
+++ b/inc/sidebars-emulator.php
@@ -65,17 +65,26 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 	 * Register all the current widgets so we can filter the get_option('widget_...') values to add instances
 	 */
 	function register_widgets() {
+		$current_url = add_query_arg( false, false );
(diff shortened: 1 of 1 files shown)


gregpriday:
Thanks for the contribution @WPprodigy! I've had a look at the code and everything looks good. I'll just do some testing, then merge.

PR #681, opened 2019-06-24
 Prevent Welcome Page Redirect During Bulk Install and TGMPA

### LINKED ISSUE #680: Redirect to a welcome page causes "headers already sent" and breaks TGM plugin activation
Hi,

Virtually all professional themes use [TGM plugin activation ](https://github.com/TGMPA/TGM-Plugin-Activation) to install required and recommended plugins. In theme guidelines is mentioned as "the only" allowed solution for required plugins installation.

Themes usually require multiple plugins.  If among them is also SO page builder, the installation process breaks at SO Page Builder plugin and does not continue.

The reason is that siteorigin panels produces "Headers already sent" error upon the installation. It is techically a php warning only, not an error, but nevertheless it breaks TGM activation class functionality. TGM stops at page builder, installs and activates it, but does not continue with other plugins.

I managed to pinpoint the error to the activation_redirect() function in siteorigin-panels.php. Specifically the exit() line. 
When the page builder plugin is installed and activated, it exits and redirects to the welcome screen. The Welcome Screen should be implemented differently, or at least find an option to avoid exiting, if TGM class is detected.

I double-tested this by just commenting out the exit() line, and loaded it through TGM as a custom source plugin (instead from WP rep). When doing this, the TGM plugin activation worked without a glitch, installing SO panels and other plugins without an issue.

### Full error trace:
```
[19-Jun-2019 11:11:06 UTC] PHP   1. {main}() /app/public/wp-admin/themes.php:0
[19-Jun-2019 11:11:06 UTC] PHP   2. require_once() /app/public/wp-admin/themes.php:10
[19-Jun-2019 11:11:06 UTC] PHP   3. do_action() /app/public/wp-admin/admin.php:253
[19-Jun-2019 11:11:06 UTC] PHP   4. WP_Hook->do_action() /app/public/wp-includes/plugin.php:465
[19-Jun-2019 11:11:06 UTC] PHP   5. WP_Hook->apply_filters() /app/public/wp-includes/class-wp-hook.php:310
[19-Jun-2019 11:11:06 UTC] PHP   6. TGM_Plugin_Activation->install_plugins_page() /app/public/wp-includes/class-wp-hook.php:286
[19-Jun-2019 11:11:06 UTC] PHP   7. TGM_Plugin_Activation->do_plugin_install() /app/public/wp-content/themes/newtheme/framework/admin/TGM/class-tgm-plugin-activation.php:761
[19-Jun-2019 11:11:06 UTC] PHP   8. TGM_Plugin_Activation->activate_single_plugin() /app/public/wp-content/themes/newtheme/framework/admin/TGM/class-tgm-plugin-activation.php:924
[19-Jun-2019 11:11:06 UTC] PHP   9. activate_plugin() /app/public/wp-content/themes/newtheme/framework/admin/TGM/class-tgm-plugin-activation.php:1077
[19-Jun-2019 11:11:06 UTC] PHP  10. do_action() /app/public/wp-admin/includes/plugin.php:699
[19-Jun-2019 11:11:06 UTC] PHP  11. WP_Hook->do_action() /app/public/wp-includes/plugin.php:465
[19-Jun-2019 11:11:06 UTC] PHP  12. WP_Hook->apply_filters() /app/public/wp-includes/class-wp-hook.php:310
[19-Jun-2019 11:11:06 UTC] PHP  13. SiteOrigin_Panels->activation_redirect() /app/public/wp-includes/class-wp-hook.php:288
[19-Jun-2019 11:11:06 UTC] PHP  14. wp_redirect() /app/public/wp-content/plugins/siteorigin-panels/siteorigin-panels.php:666
[19-Jun-2019 11:11:06 UTC] PHP  15. header() /app/public/wp-includes/pluggable.php:1251
```

### How to replicate the issue:
Theme should require class-tgm-plugin-activation.php and load some random plugins:
here is a ready-to-use code to add into the functions.php (you only need class-tgm-plugin-activation.php in the theme folder):

```
require_once get_template_directory() . 'class-tgm-plugin-activation.php';
add_action( 'tgmpa_register', 'test_install_plugins' );

function test_install_plugins() {
    $plugins = array(
        array(
            'name' => 'Contact form 7',
            'slug' => 'contact-form-7',
        ),
        array(
            'name'      => 'Page Builder by SiteOrigin',
            'slug'      => 'siteorigin-panels',
            'required'  => true,
        ),           
        array(
            'name'      => 'SiteOrigin Widgets Bundle',
            'slug'      => 'so-widgets-bundle',
            'required'  => true,
        ),        
        array(
            'name' => 'Newsletter',
            'slug' => 'newsletter',
        ),
        array(
            'name' => 'Safe SVG',
            'slug' => 'safe-svg',
        ),
    );
    $config = array(
        'menu'         => 'install-required-plugins',
        'is_automatic' => true,
    );
    tgmpa($plugins, $config);
}
```

To see the warning,, set WP_DEBUG to true in wp-config.php
To trace everything, you can add this snippet:

```
define('WP_DEBUG', true);
ini_set('log_errors',TRUE);
ini_set('error_reporting', E_ALL);
ini_set('error_log', dirname(__FILE__) . '/error_log.txt');
ini_set('display_errors',1);
ini_set('display_startup_errors',1);
error_reporting(-1);
```

AlexGStapleton:
Resolves #680

When installing using TGMPA, the redirect will happen on the next page the user views and not directly after activation.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -675,13 +675,11 @@ class SiteOrigin_Panels {
 		if ( get_transient( 'siteorigin_panels_activation_welcome' ) ) {
 
 			// Postpone redirect in certain situations
-			if ( wp_doing_ajax() || is_network_admin() || isset( $_GET['activate-multi'] ) ) {
-				return false;
+			if ( ! wp_doing_ajax() && ! is_network_admin() && ! isset( $_GET['activate-multi'] ) ) {
+				delete_transient( 'siteorigin_panels_activation_welcome' );
+				wp_safe_redirect( admin_url( 'options-general.php?page=siteorigin_panels#welcome' ) );
+				exit();
 			}
-
-			delete_transient( 'siteorigin_panels_activation_welcome' );
-			wp_safe_redirect( admin_url( 'options-general.php?page=siteorigin_panels#welcome' ) );
-			exit();
 		}
 	}
 }
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -38,7 +38,8 @@ class SiteOrigin_Panels {
 		add_filter( 'siteorigin_panels_data', array( $this, 'process_panels_data' ), 5 );
 		add_filter( 'siteorigin_panels_widget_class', array( $this, 'fix_namespace_escaping' ), 5 );
 		
-		add_action( 'activated_plugin', array($this, 'activation_redirect') );
+		add_action( 'activated_plugin', array($this, 'activation_flag_redirect') );
+		add_action( 'admin_init', array($this, 'activation_do_redirect') );
 
 		if (
 			is_admin() ||
@@ -657,19 +658,30 @@ class SiteOrigin_Panels {
 	}
 	
 	/**
-	 * Redirect to a welcome page after activation.
+	 *  Flag redirect to welcome page after activation
 	 *
 	 * @param $plugin
 	 */
-	public function activation_redirect( $plugin ){
-		if( $plugin == plugin_basename( __FILE__ ) ) {
+	public function activation_flag_redirect( $plugin ) {
+		if ( $plugin == plugin_basename( __FILE__ ) ) {
+			set_transient( 'siteorigin_panels_activation_welcome', true, 30 );
+		}
+	}
 
-			// Prevent redirect in certain situations
+	/**
+	 * Redirect to a welcome page after activation.
+	 */
+	public function activation_do_redirect() {
+		if ( get_transient( 'siteorigin_panels_activation_welcome' ) ) {
+
+			// Postpone redirect in certain situations
 			if ( wp_doing_ajax() || is_network_admin() || isset( $_GET['activate-multi'] ) ) {
 				return false;
 			}
-			
-			exit( wp_redirect( admin_url( 'options-general.php?page=siteorigin_panels#welcome' ) ) );
+
+			delete_transient( 'siteorigin_panels_activation_welcome' );
+			wp_safe_redirect( admin_url( 'options-general.php?page=siteorigin_panels#welcome' ) );
+			exit();
 		}
 	}
 }
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -38,7 +38,8 @@ class SiteOrigin_Panels {
 		add_filter( 'siteorigin_panels_data', array( $this, 'process_panels_data' ), 5 );
 		add_filter( 'siteorigin_panels_widget_class', array( $this, 'fix_namespace_escaping' ), 5 );
 		
-		add_action( 'activated_plugin', array($this, 'activation_redirect') );
+		add_action( 'activated_plugin', array($this, 'activation_flag_redirect') );
+		add_action( 'admin_init', array($this, 'activation_do_redirect') );
 
 		if (
 			is_admin() ||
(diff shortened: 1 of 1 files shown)


braamgenis:
It seems like we could avoid the issue by just adding a check for bulk activation. I.e. with everything as it was, just add a check for `$_POST['action'] == 'activate-selected'` in the `activation_redirect` method. This keeps things nice and simple with less potential for bugs.

@AlexGStapleton Could you confirm whether just adding the condition resolves the issue with TGM too?

AlexGStapleton:
That technically resolve this issue. as completely it prevents the welcome redirect - there won't be any redirect. Is that okay? 

I was under the impression that the Welcome redirect should be maintained, just delayed to avoid this issue.

braamgenis:
> completely it prevents the welcome redirect

It should only prevent the redirect for bulk activation. I.e. when a user selects multiple plugins on the plugins page and then selects _Activate_ from the _Bulk Actions_ dropdown. We definitely don't want the redirect with bulk activations as there may be other plugins trying to redirect too.

AlexGStapleton:
I'm sorry I wasn't clear. By _completely_ I meant that no redirect will ever occur. This PR allows for a delayed redirect. Checking `$_POST['action'] == 'activate-selected'` in `activation_redirect` will prevent the redirect entirely - no redirect will ever occur.

Does that make sense?

braamgenis:
> will prevent the redirect entirely

It only prevents the redirect for _**bulk**_ activations.

To test, revert all the changes in this PR and replace the `activation_redirect` with this:
```
public function activation_redirect( $plugin ){
	$is_bulk = ! empty( $_POST['action'] ) && $_POST['action'] == 'activate-selected';
	if( ! $is_bulk && $plugin == plugin_basename( __FILE__ ) ) {
		exit( wp_redirect( admin_url( 'options-general.php?page=siteorigin_panels#welcome' ) ) );
	}
}
```

Now when you activate PB along with one or more other plugins, there _**won't**_ be a redirect, but when you activate PB on it's own, the redirect should still occur.

AlexGStapleton:
Yes, sorry - I had accidentally omitted bulk from that sentence.

I was under the impression that the Welcome redirect should be maintained with bulk redirects, just delayed to avoid this issue - this is akin to what WooCommerce does. Is that not the case?

AlexGStapleton:
The last commit completely prevents the redirect after bulk action rather than delaying it.

siteorigin-panels.php
```diff
@@ -663,6 +663,12 @@ static function get_widget_instance( $class_or_hash ) {
 	 */
 	public function activation_redirect( $plugin ){
 		if( $plugin == plugin_basename( __FILE__ ) ) {
+
+			// Prevent redirect in certain situations
+			if ( wp_doing_ajax() || is_network_admin() || isset( $_GET['activate-multi'] ) ) {
```
braamgenis: This doesn't work for me. In which circumstances is `activate-multi` used? I don't see it when activating multiple plugins from the plugins page.

As far as I can tell, WP sets `$_POST['action']` to `'activate'` when a single plugin is activated and sets it to `'activate-selected'` when multiple plugins are activated.

siteorigin-panels.php
```diff
@@ -663,6 +663,12 @@ static function get_widget_instance( $class_or_hash ) {
...
+				return false;
```
braamgenis: This shouldn't be returning anything. It's an action handler, not a filter.

AlexGStapleton: Oh whoops. Fixing now.

siteorigin-panels.php
AlexGStapleton: >  I don't see it when activating multiple plugins from the plugins page.

That's when `activate-multi` is used.  I'm not too sure why it's not triggering for you but it's definitely meant to. [Here's where WordPress uses this](https://github.com/WordPress/WordPress/blob/master/wp-admin/plugins.php#L126). Does disabling all non-SO plugins help? (there might be something interfering)

AlexGStapleton: I've identified why this is happening. [This redirect](https://github.com/WordPress/WordPress/blob/master/wp-admin/plugins.php#L126) is done _after_ the plugins are activated. This results in `activate-multi` not existing until after  the `activated_plugin` hook is used.

I can't see any other method of preventing this without using two hooks (`activated_plugin` `and admin_init` - this can be seen in my first commit). Do you see any other ways?

braamgenis: Does my previous suggestion not work for you?

AlexGStapleton: Sorry for the delay - lost track of this reply.

It doesn't, no. I applied the suggested change to a dev version and a release version (to rule out an issue with the dev version), and it didn't work for either.

Did it work for you?

braamgenis: Sorry about the runaround, I must have missed something in my initial checking. I checked the fix using a transient again and that's working for me now. :confounded:  SO I think reverting to that is the best way to go for this.

AlexGStapleton: No worries. I think there was something wrong with my previous dev install as the current commit seriously did work so I guess everything evens out.

Just reverted the last commit.

siteorigin-panels.php
AlexGStapleton: Just actually fixed this.

Issue #269, opened 2016-07-28
Add Responsive fields for Row Bottom Margin


AlexGStapleton:
This is a semi-common indirect request (meaning users ask questions that would have been resolved with this functionality) but users would like the ability to adjust row bottom margin for mobile and tablet devices.

This setting should also have these responsive fields in the Page Builder settings.

It should be noted that I'm not directly suggesting dedicated fields for things like row gutter as this is purely a vertical spacing issue that I would like to resolve.

gregpriday:
@AlexGStapleton we do now have a setting for responsive row padding. Technically a user could add an additional bottom padding for mobile devices. This is slightly different because this bottom padding will be added to the row styling.

Do you think responsive bottom padding is enough, or is there still a need for responsive bottom margin?

AlexGStapleton:
In my opinion, I think we should. With that said, we could get away with having a simple checkbox for "remove bottom margin on mobile" if it makes it easier.

I think a more important issue is the lack of control over widget and cell (only notice on mobile devices) bottom margin.

AlexGStapleton:
Actually, I change my stance from "we should" to be "there are more reasons to do this than to not do so". Adjusting the padding for additional spacing (or to remove spacing) isn't really an option when you have a background.

Misplon:
Ref: https://github.com/siteorigin/siteorigin-panels/tree/feature/mobile-styles

PR #687, opened 2019-08-02
Mobile styles

gregpriday:
Resolves GH-269

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -64,14 +64,17 @@ class SiteOrigin_Panels_Renderer {
 		$settings = siteorigin_panels_setting();
 		$panels_tablet_width = $settings['tablet-width'];
 		$panels_mobile_width = $settings['mobile-width'];
-		$panels_margin_bottom = $settings['margin-bottom'];
-		$panels_mobile_margin_bottom = $settings['margin-bottom'];
 		$panels_margin_bottom_last_row = $settings['margin-bottom-last-row'];
 
 		$css = new SiteOrigin_Panels_Css_Builder();
 
 		$ci = 0;
 		foreach ( $layout_data as $ri => $row ) {
+			
+			// Filter the bottom margin for this row with the arguments
+			$panels_margin_bottom = apply_filters( 'siteorigin_panels_css_row_margin_bottom', $settings['margin-bottom'] . 'px', $row, $ri, $panels_data, $post_id );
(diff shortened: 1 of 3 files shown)


inc/renderer.php
```diff
@@ -133,101 +137,105 @@ public function generate_css( $post_id, $panels_data = false, $layout_data = fal
...
-					'margin-bottom' => apply_filters( 'siteorigin_panels_css_row_margin_bottom', $panels_margin_bottom . 'px', $row, $ri, $panels_data, $post_id )
+					'margin-bottom' => $panels_margin_bottom
 				) );
 			}
 
 			$collapse_order = ! empty( $row['style']['collapse_order'] ) ? $row['style']['collapse_order'] : ( ! is_rtl() ? 'left-top' : 'right-top' );
 
-			if ( $settings['responsive'] ) {
-
+			if ( $settings['responsive'] && ! empty( empty( $row['style']['collapse_behaviour'] ) ) ) {
```
braamgenis: This is pretty hard to read and seems unnecessary. `! empty( empty( $var ) )` should be the same as just `empty( $var )`?

inc/renderer.php
```diff
@@ -133,101 +137,105 @@ public function generate_css( $post_id, $panels_data = false, $layout_data = fal
...
-							$remove_bottom_margin .= 'last-child(' . ( count( $row['cells'] ) % 2 == 0 ? '-n+2' : '1' ) . ')';
-						} else {
-							$remove_bottom_margin .= 'child(-n+2)';
-						}
+					$css->add_cell_css( $post_id, $ri, false, '', array(
+						'-ms-flex'      => '0 1 50%',
+						'-webkit-flex'  => '0 1 50%',
+						'flex'          => '0 1 50%',
+						'margin-right'  => '0',
+						'margin-bottom' => $panels_margin_bottom,
```
braamgenis: As far as I can tell, this is the only line that changed to remove the `'px` unit. That correct? Not sure why the diff marked so much as changed around this line. 😕

inc/renderer.php
gregpriday: Aaah yeah, that's a copy paste typo. Will fix it up.

gregpriday:
Yeah, the diff got a bit crazy because I combined 2 if statements that shouldn't have been separate. I'll fix up the double empty.

PR #690, opened 2019-08-15
Release 2.10.7

gregpriday:
* Added setting for mobile specific margin.
* Prevent Welcome Page Redirect During Bulk Install and TGMPA
* Added support for password settings field.
* Layout Block: Add filter to control whether Add Layout Block button is shown or not.
* Fixed issue with widget duplication after moving a widget.
* Fixed Read More Custom Text issue.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -673,6 +673,7 @@ class SiteOrigin_Panels {
 	 */
 	public function activation_do_redirect() {
 		if ( get_transient( 'siteorigin_panels_activation_welcome' ) ) {
+			delete_transient( 'siteorigin_panels_activation_welcome' );
 
 			// Postpone redirect in certain situations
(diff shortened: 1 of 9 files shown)


siteorigin-panels.php
```diff
@@ -327,7 +328,7 @@ public function generate_post_content( $content ) {
 						$content = explode( $matches[0], $content, 2 );
 						$content = $content[0];
 						$content = force_balance_tags( $content );
-						if ( ! empty( $matches[1] ) && ! empty( $more_link_text ) ) {
+						if ( ! empty( $matches[1] ) ) {
```
braamgenis: Took a little while to follow what was happening here, so some comments about how it's trying to figure out what to display as the "Read more" text would be nice, but obviously not a blocker.

PR #696, opened 2019-08-29
Add support for widgets that don't have a dedicated class

gregpriday:
The change https://github.com/siteorigin/siteorigin-panels/commit/2a6179a04ac88ce712e49bae6bde7a854b050c54 introduced an issue where widgets that didn't have a dedicated class weren't showing up in Page Builder. This PR fixes that by only using the `$widget_obj` class, when the given `$class` from the array key is a hash.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -751,8 +751,8 @@ class SiteOrigin_Panels_Admin {
 	function get_widgets() {
 		global $wp_widget_factory;
 		$widgets = array();
-		foreach ( $wp_widget_factory->widgets as $widget_obj ) {
-			$class = get_class( $widget_obj );
+		foreach ( $wp_widget_factory->widgets as $class => $widget_obj ) {
+			$class = preg_match( '/[0-9a-f]{32}/', $class ) ? get_class( $widget_obj ) : $class;
 			$widgets[ $class ] = array(
 				'class'       => $class,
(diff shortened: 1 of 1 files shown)


gregpriday:
@braamgenis this is a really minor change that seems to fix the issue we were having with plugins like Codelights. I just need a sanity check that this isn't going to break anything in some other situation.

inc/admin.php
```diff
@@ -751,8 +751,8 @@ function update_home_on_theme_change() {
 	function get_widgets() {
 		global $wp_widget_factory;
 		$widgets = array();
-		foreach ( $wp_widget_factory->widgets as $widget_obj ) {
-			$class = get_class( $widget_obj );
+		foreach ( $wp_widget_factory->widgets as $class => $widget_obj ) {
+			$class = preg_match( '/[0-9a-f]{32}/', $class ) ? get_class( $widget_obj ) : $class;
```
braamgenis: The only question is whether the hashes are guaranteed, currently and in future, to be 32 character strings. E.g. what happens if they decide to start using `spl_object_id` (available in PHP 7.2) instead of `spl_object_hash` to generate the hashes? And of course, the fairly unlikely, but still valid case of having a class name containing only the above characters and having a length of 32. Love PHP. ❤️ 🤷‍♂

gregpriday: From a little bit of testing and a few examples online, it seems like the hash is always 32 character hex.

If WordPress or PHP change anything in the future, then we can just expand the regex to pick that up. For now, it's a lot more robust than trying to identify something that looks like a class name.

PR #708, opened 2019-10-21
Resolve Issue Caused By locate_template Preventing Plugins from Adding Post Loop Templates

### LINKED ISSUE #707: Post Loop Templates: Accomodate templates located in plugins
We introduced https://github.com/siteorigin/siteorigin-panels/commit/c1750752122b9813b9f53c1c570c73e867ed7591 to make it easier for child themes to override parent templates. Unfortunately, this change broke https://github.com/siteorigin/siteorigin-panels/commit/91c9bb13b8a96bd21d03e6ef847b4e430bbb5135 which allowed templates to be loaded from child themes. 

Example:

```
function so_custom_postloop_directories( $directories ) {
	$directories[] = plugin_dir_path( __FILE__ ) . '/loops/';

	return $directories;
}
add_filter( 'siteorigin_panels_postloop_template_directory', 'so_custom_postloop_directories' );

```
We need a solution that fixes `siteorigin_panels_postloop_template_directory` but doesn't impact users that have moved templates into child themes to override them.

AlexGStapleton:
This PR resolves #707.

### DIFF
--- a/inc/widgets/post-loop-helper.php
+++ b/inc/widgets/post-loop-helper.php
@@ -17,7 +17,6 @@ class SiteOrigin_Panels_Widgets_PostLoop_Helper extends SiteOrigin_Widget {
 		$template_options = array();
 		if( ! empty( $templates ) ) {
 			foreach( $templates as $template ) {
-				$headers = get_file_data( locate_template( $template ), array(
 				// Is this template being added by a plugin?
 				$filename = locate_template( $template );
 				if ( ! $filename ) {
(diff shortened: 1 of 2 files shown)


Misplon:
From my end, this PR is functional. 
Loops are automatically overridden when inserted into child themes. 
Within plugins, `siteorigin_panels_postloop_templates` is functioning as it should.

Example:
```
function so_postloop_template_test( $templates ) {
	$templates[] = 'so-custom-post-loop/loops/loop-grid.php';

	return $templates;
}
add_filter( 'siteorigin_panels_postloop_templates', 'so_postloop_template_test' );
```

Issue #702, opened 2019-10-09
iOS: Resolve heading sizing


Misplon:
![IMG_2165](https://user-images.githubusercontent.com/789159/66516089-92530900-eae0-11e9-9d0f-0ce29c24877c.png)

Section, heading and label sizing isn't what it should be in iOS.

PR #714, opened 2019-12-09
Prevent WebKit font size adjustments

Misplon:
Prevent WebKite font size adjustments in the widget modal.

Resolves https://github.com/siteorigin/siteorigin-panels/issues/702.


Misplon:
@AlexGStapleton Quick check when able. Thanks.

Misplon:
Super, thanks for checking.

Issue #716, opened 2020-01-08
Layout Block: Harden against styling impacts from Twenty Twenty


Misplon:
<img width="851" alt="01" src="https://user-images.githubusercontent.com/789159/71977736-3c4a4a80-3222-11ea-8b4e-9aa8ae1fe1b9.png">
<img width="835" alt="02" src="https://user-images.githubusercontent.com/789159/71977737-3c4a4a80-3222-11ea-977f-5e727b3351d3.png">

PR #717, opened 2020-01-08
Twenty twenty fixes

Misplon:
Resolves https://github.com/siteorigin/siteorigin-panels/issues/716. 

Perhaps there is a better way to achieve superior specificity? `text-decoration: none;` is the only change where I'm not repeating rules. The link color and font family changes are just re-adding rules further down the cascade.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -1971,6 +1971,7 @@
 		line-height: 1.4em;
 
 		ul {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
You could apply the `text-decoration;` using [this selector](https://github.com/siteorigin/siteorigin-panels/blob/6c8ed87e3a24b2b692173396bcc23ca429b81594/css/admin.less#L686).

Misplon:
Done, thanks.

Issue #733, opened 2020-03-27
panels_data not available within Live editor - PHP Notice


andrejarh:
In backend, when using live editor
I am shown a notice instead of the content when trying to load panels data in the theme banner area. 

**"Notice: Undefined variable: panels_data in /home/wp-content/plugins/siteorigin-panels/inc/renderer.php on line 658"**

The theme banner area renders siteorigin content before the wp main query (just below header). It displays page builder panel data by using "echo siteorigin_panels_render()" function. 

So the notice is displayed for page builder content which is not part of currently editable post content and is outside the main WP loop.

![image](https://user-images.githubusercontent.com/10053621/77745130-0ba4d080-701b-11ea-8ef2-e7f3777b45b7.png)

On the front end, the banner area displays without problems. Only live editor displays the PHP notice.

Ideally the content should display in live editor as well, but should not be editable for the obvious reasons.

AlexGStapleton:
Hi @andrejarh,

Can you please clarify which theme of yours this is happening with? This will allow me to run some tests using a setup that we know can result in this issue. 

Kind regards,
Alex

andrejarh:
Hi Alex,
To isolate the issue without additional clutter, I created a Vantage child theme. 
The child theme creates a banner area in SiteOrigin Vantage theme, a special post type and a widget for displaying its content.

Please download the Vantage child theme here: https://github.com/andrejarh/vantage-test-child-theme.  Demo page import file is also available, to demonstrate the Live Editor issue.

The issue https://github.com/siteorigin/siteorigin-panels/issues/667 seem to also be connected to this one and can be observed on the demo as well.

Please check it out. All feedback is welcome.
Andrej

AlexGStapleton:
Hi @andrejarh,

Thank you for the child theme. I've been able to replicate this issue and I've submitted a #734 that will resolve it. 

@Misplon In addition to the steps outlined in @andrejarh's child theme, you'll need to edit the Page called "A page with banner area" and update "test_static_block_meta_box" with the id of imported test page.

Kind regards,
Alex

AlexGStapleton:
Resolved by https://github.com/siteorigin/siteorigin-panels/pull/734

PR #734, opened 2020-03-30
Prevent Live Editor Notice that can appear when using SO Renderer outside of main loop

### LINKED ISSUE #667: siteorigin_panels_render() not rendering anymore in Live editor
I have a custom widget and a custom post type. 
The widget just renders the content of a (custom post-type) post. 

To display the content, the widget uses this line of code: 
echo siteorigin_panels_render($custom_post_id);

 **what does the widget do** 
The idea behind the widget is to simplify (repeatable) content updates.
Let's say that the same content (rows + widgets) is displayed on 1000 different pages, then you need to add a small change to it. If the content is edited on a specific custom post type post, You only need to change it there and the content automatically updates on all 1000 pages... 

**pinpointing the issue**
It works perfectly on the front end. However in the live editor, the siteorigin_panels_render() doesn't render anymore in the Live editor using the latest SO panels versions. 
I tested it on an old version 2.10.1, and can confirm it is still working there. Also on all older versions.

I tried to pinpoint the issue, so i compared the new and the old versions. I did not notice any  changes within the siteorigin_panels_render() or SiteOrigin_Panels::renderer() function. 
I don't quite understand why the content isn't displayed anymore.

**the question**
So my question is: Is there a way to display a specific post content within the live editor? Is there some other function than siteorigin_panels_render(), which I should use?

Thanks!

AlexGStapleton:
Resolve #667 #733

Please refer to #733 for a child theme that will allow you to replicate this issue - it's possible to replicate this issue without it, it's just simplier.

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -648,7 +648,9 @@ class SiteOrigin_Panels_Renderer {
 
 				$panels_data = ! empty( $layouts[ $prebuilt_id ] ) ? $layouts[ $prebuilt_id ] : current( $layouts );
 			}
-		} else {
+		}
+
+		if ( ! empty( $post_id ) && empty( $panels_data ) ) {
 			if ( post_password_required( $post_id ) ) {
 				return false;
 			}


Misplon:
I've replicated the steps requested and can confirm the notice issue is resolved.

PR #740, opened 2020-04-04
Live editor: Use Global Settings for device widths and prevent scaling

### LINKED ISSUE #284: Live Editor UX: Force device view iframe at actual width rather than being responsive
When using the live editor on an iPad (when in vertical orientation) all three views will look the same due to the iframe being responsive (max-width vs width). We should show the iframe at a static width rather than a dynamic one to provide a consistent preview overall. If this width exceeds the browser width, we should overflow scroll it.

AlexGStapleton:
Resolve #284

This PR changes the device resolutions in the Live Editor to reflect the global Page Builder Settings. It also prevents the resizing of the preview to allow for accurate previews on devices that use a resolution higher than the one they're previewing.

To prevent the misalignment of the preview at different resolutions and adjusted device widths (if the user changes to a device other than the default), how the preview is displayed had to modified to be closer to rely less on negative margins to account for the Live Editor side. Instead, a calculated width was used instead.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -2175,7 +2175,6 @@
 		top: 0;
 		right: 0;
 		bottom: 0;
-		left: @live_editor_width;
 		background-color: #191e23;
 
 		form {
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
@Misplon What are your thoughts on animating the resize of the preview? For an example of how this would look, open the customizer and change the device.

Misplon:
Adding an animation the same as the Customizer would be good.

PR #749, opened 2020-05-08
Fix preview issue when using Live Editor before saving

### LINKED ISSUE #748: Live editor bug (new pages)
1. When a new page is created and opened with live editor (before publishing the page), The live editor does not work. Any changes to the site are not displayed. 

Also a thing I noticed is that WP Admin bar is displayed in the live editor, when editing drafts (which is not the case when editing already saved posts).

This was tested on a fresh installation with the latest official WP version, default Twentytwenty WP theme and Page builder version 2.10.16. No third party plugins were installed. 

2. Additionally when I installed the widget bundle, and added the Features widget. I got this error:

Screenshot 2020-05-07 at 09 59 15

AlexGStapleton:
Resolves #748

@Misplon Please confirm the test.html file I sent you the other day over Slack still works as expected for you.

### DIFF
--- a/inc/live-editor.php
+++ b/inc/live-editor.php
@@ -23,20 +23,21 @@ class SiteOrigin_Panels_Live_Editor {
 
 	public function xss_headers(){
 		global $post;
-		if(
-			! empty( $_POST['live_editor_panels_data'] ) &&
-			! empty( $post->ID ) &&
-			current_user_can( 'edit_post', $post->ID ) &&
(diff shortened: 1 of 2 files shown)


Misplon:
I can't recreate the original issue for some reason but I did test security in this branch. I also did a build and tested on another localhost site and all seems the same as in master so I think we're still good on that front.

Misplon:
I can recreate the original issue. This branch seems to resolve it and the console errors.

Issue #755, opened 2020-05-25
Legacy Layout Engine: Ensure IE 11 is detected as old browser 


AlexGStapleton:
A few users in the past have reported that the Legacy Layout Engine Detect Older Browsers setting isn't identifying IE 11. This is a legacy browser and it requires the legacy rendering so we should ensure it's being properly detected.

PR #756, opened 2020-05-25
Legacy Layout: Detect IE 11 as legacy browser

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/755

This PR changes the legacy device detection to include effectively [all versions of IE](https://www.whatismybrowser.com/guides/the-latest-user-agent/internet-explorer). Edge does not include Trident in its user agent so it won't be picked up by this change.


Misplon:
A forum user tested for us, it's working.

Issue #726, opened 2020-02-12
Add content analysis integration for Rank Math


Misplon:
Look at adding content analysis integration for [Rank Math](https://rankmath.com/kb/content-analysis-api/). Currently active installs at 200k.

Misplon:
https://rankmath.com/kb/content-analysis-api/

PR #752, opened 2020-05-13
Improved Yoast Comatability

gregpriday:
This PR fixes our live Yoast integration in the classic editor. It also resolves https://github.com/siteorigin/siteorigin-panels/issues/675 and https://github.com/siteorigin/siteorigin-panels/issues/647

PR #767, opened 2020-06-17
Add content analysis integration for Rank Math 

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/726

While https://github.com/siteorigin/siteorigin-panels/pull/752 indirectly helped, there were certain widget not being picked up correctly. Widgets such as the Image widget. To verify this PR helps, please add a focus keyword and then add an image with that focus keyword in the alt. In develop, the `Focus Keyword found in image alt attribute(s)` check will never pass.

I went with renaming yoast-compat.js rather than creating a new dedicated rankmath-comat.js file due to the vast amount of code that would need to be repeated.

Screenshot_2020-06-18 Edit Page ‹ SiteOrigin — WordPress

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -536,7 +536,7 @@ class SiteOrigin_Panels_Admin {
 					SITEORIGIN_PANELS_VERSION,
 					true
 				);
-			} else if ( defined( 'RANK_MATH_VERSION' ) && wp_script_is( 'rank-math-analyzer' ) ) {
+			} elseif ( defined( 'RANK_MATH_VERSION' ) && wp_script_is( 'rank-math-analyzer' ) ) {
 				wp_enqueue_script(
 					'so-panels-seo-compat',
 					siteorigin_panels_url( 'js/seo-compat' . SITEORIGIN_PANELS_JS_SUFFIX . '.js' ),
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -70,8 +70,8 @@ class SiteOrigin_Panels_Admin {
 
 
 		// Enqueue Yoast compatibility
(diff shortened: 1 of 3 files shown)


inc/admin.php
```diff
@@ -526,16 +526,26 @@ function enqueue_admin_scripts( $prefix = '', $force = false ) {
...
+		if ( self::is_admin() ) {
+			if ( defined( 'WPSEO_FILE' ) && wp_script_is( 'yoast-seo-metabox' ) ) {
+				wp_enqueue_script(
+					'so-panels-seo-compat',
+					siteorigin_panels_url( 'js/seo-compat' . SITEORIGIN_PANELS_JS_SUFFIX . '.js' ),
+					array('jquery', 'yoast-seo-metabox' ),
+					SITEORIGIN_PANELS_VERSION,
+					true
+				);
+			} else if ( defined( 'RANK_MATH_VERSION' ) && wp_script_is( 'rank-math-analyzer' ) ) {
```
Misplon, line 539: Probably best to switch to `elseif` to keep with the standard https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#use-elseif-not-else-if.

js/siteorigin-panels/view/builder.js
```diff
@@ -775,16 +775,19 @@ module.exports = Backbone.View.extend( {
 			contentEd.fire( 'keyup' );
 		}
 
-		this.triggerYoastSeoChange();
+		this.triggerSeoChange();
 	},
 
 	/**
-	 * Trigger a change on Yoast SEO
+	 * Trigger a change SEO plugins
```
Misplon, line 782: Can we perhaps change this to: `Trigger a change in SEO plugins.`

Issue #723, opened 2020-02-02
Provide list of used images to Yoast Sitemap


AlexGStapleton:
We should provide Yoast Sitemap with a list of the images added using SiteOrigin Page Builder and the SiteOrigin Layouts block. This can be done using the wpseo_xml_sitemap_img filter.

https://kb.yoast.com/kb/images-in-the-xml-sitemap/

PR #769, opened 2020-07-02
Yoast Sitemap: List images added using Page Builder

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/723

I used DOMDocument rather than regex as Yoast changed from regex to DOMDocument. It appears to be due to reliability issues but I wasn't able to find concrete information on this.

### DIFF
--- a/compat/yoast-sitemap.php
+++ b/compat/yoast-sitemap.php
@@ -23,13 +23,11 @@ function siteorigin_yoast_sitemap_images_compat( $images, $post_id ) {
 		foreach ( $dom->getElementsByTagName( 'img' ) as $img ) {
 			$src = $img->getAttribute( 'src' );
 
-			if ( empty( $src ) || $src !== esc_url( $src ) ) {
-				continue;
+			if ( ! empty( $src ) && $src == esc_url( $src ) ) {
+				$images[] = array(
+					'src'   => $src,
+				);
 			}
-
-			$images[] = array(
-				'src'   => $src,
-			);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


compat/yoast-sitemap.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+function siteorigin_yoast_sitemap_images_compat( $images, $post_id ) {
+	if ( get_post_meta( $post_id, 'panels_data', true ) && class_exists( 'DOMDocument' ) ) {
+		$content = SiteOrigin_Panels::renderer()->render(
+			$post_id,
+			false
+		);
+
+		libxml_use_internal_errors( true );
+		$dom = new DOMDocument();
```
gregpriday, line 19: If a user doesn't have `php-xml` installed, then this will cause a fatal error. I don't think it's very common, but I remember encountering this back in my early days of theme dev. It's worth throwing in a `class_exists( 'DOMDocument' )` check at the start of `siteorigin_yoast_sitemap_images_compat`.

That said, maybe Yoast is already checking for this, so a user will never hit the `wpseo_sitemap_urlimages` filter if they don't have `php-xml` installed.

AlexGStapleton: Yoast checks if `DOMDocument` is enabled (if it's not, it'll never trigger our code), but they're not checking if `php-xml` is. Their usage is significantly before ours so if there is a fatal error, it'll be with their code first. It would be considering adding a `php-xml` check in the offchance they add a check also.

Is your preference with checking with `extension_loaded` or will `function_exists` suffice?

gregpriday: Either would be good here.

Issue #765, opened 2020-06-14
Add keyboard navigation to Page Builder interface


AlexGStapleton:
A recent accessibility review found that users aren't able to navigate the SiteOrigin Page Builder interface with just a keyboard. We should add tab indexes and make general CSS changes to apply `:hover` CSS to `:focus` to rectify this.

PR #772, opened 2020-07-06
Improve Page Builder Interface accessibility

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/765

Status (pending approval):
- [x] Navigating page builder rows, and columns
- [x] Add Widgets view
- [x] Layouts view
- [x] History
- [x] Live Editor
- [x] Row/Widget interface navigation
- [x] Widget contents navigation https://github.com/siteorigin/so-widgets-bundle/pull/1206

Note:
- There's no way to reorder widgets or rows. This needs further investigation and will be implemented in a future PR.

### DIFF
--- a/js/siteorigin-panels/helpers/accessibility.js
+++ b/js/siteorigin-panels/helpers/accessibility.js
@@ -6,7 +6,7 @@ module.exports = {
 	 */
 	triggerClickOnEnter: function( e ) {
 		if ( e.which == 13 ) {
-			$(e.target).trigger('click');
+			$( e.target ).trigger( 'click' );
 		}
 	},
 
--- a/js/siteorigin-panels/helpers/accessibility.js
+++ b/js/siteorigin-panels/helpers/accessibility.js
@@ -1,0 +1,13 @@
+var $ = jQuery;
+
+module.exports = {
+	/**
+	 * Trigger click on valid enter key press.
+	 */
+	triggerClickOnEnter: function( e ) {
+		if ( e.which == 13 ) {
+			$(e.target).trigger('click');
+		}
+	},
+
+};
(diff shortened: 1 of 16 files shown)


js/siteorigin-panels/view/builder.js
```diff
@@ -247,9 +249,13 @@ module.exports = Backbone.View.extend( {
 
 		// Switch back to the standard editor
 		if ( this.supports( 'revertToEditor' ) ) {
-			metabox.find( '.so-switch-to-standard' ).click( function ( e ) {
+			metabox.find( '.so-switch-to-standard' ).on( 'click keyup', function ( e ) {
```
Misplon, line 252: Perhaps changed to `function( e )`

js/siteorigin-panels/helpers/accessibility.js
```diff
@@ -0,0 +1,13 @@
+var $ = jQuery;
+
+module.exports = {
+	/**
+	 * Trigger click on valid enter key press.
+	 */
+	triggerClickOnEnter: function( e ) {
+		if ( e.which == 13 ) {
+			$(e.target).trigger('click');
```
Misplon, line 9: Perhaps `$( e.target ).trigger( 'click' );`

rockstarrossi:
Yes

AlexGStapleton:
@Misplon The last commit will handle the widget and layouts views.

Some thoughts:

Can you think of a way to trigger the context menu? We could use a button press, but how would we communicate that? It really needs to be obvious.

How do you want to handle the live editor? I honestly can't think of a way for us to reliabily get it working as we can't always rely on the theme to handle things in a consistent manner.

Misplon:
Discussion was had via Slack.

Misplon (approved):

gregpriday (approved):

Issue #737, opened 2020-04-02
AMP plugin breaking Text widget


Misplon:
The [AMP plugin](https://wordpress.org/plugins/amp/) is breaking the Text widget in Page Builder. The SiteOrigin Editor, Visual Editor, and Custom HTML widgets aren't impacted. To replicate the issue, install and activate the AMP plugin, insert a Text widget into a page and open it. The widget fails to load correctly. No console or PHP errors were present when I tested.

PR #773, opened 2020-07-06
Resolve AMP Editing Issue

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/737

The AMP plugin overrides the text widget and extends the base text widget class to do this ([reference](https://github.com/ampproject/amp-wp/blob/6935c0c015e52fc58b27c4790348bae56522c875/includes/widgets/class-amp-widget-text.php)). The text widget requires its JS to be directly output so this PR adds the AMP widget as a Core JS widget.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -83,11 +83,6 @@ class SiteOrigin_Panels_Admin {
 				add_filter( 'display_post_states', array( $this, 'add_panels_post_state' ), 10, 2 );
 			}
 		}
-
-		// Compatibility with AMP plugin
-		if ( function_exists( 'amp_bootstrap_plugin' ) ) {
-			require_once plugin_dir_path( __FILE__ ) . '../compat/amp.php';
-		}
 	}
 
 	/**
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -1024,14 +1024,14 @@ class SiteOrigin_Panels_Admin {
 	function is_core_js_widget( $widget ) {
 		$js_widgets = apply_filters(
 			'siteorigin_panels_core_js_widgets',
-				array(
-					'WP_Widget_Custom_HTML',
-					'WP_Widget_Media_Audio',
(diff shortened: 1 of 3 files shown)


Misplon:
Awesome, thanks.

When switching to this branch I'm seeing an existing Text widget as missing.

Misplon:
A new Text widget is fine. Alex, can you see if you're getting the same result with existing Text widgets?

AlexGStapleton:
Unfortunately, that's the AMP plugins doing. It removes the base widget in favor of its version of the widget. I could add some code to re-add it, but it will lose the modifications that AMP implments.

The base text widget is already accounted for [here](https://github.com/siteorigin/siteorigin-panels/pull/773/files#diff-7199c031b61520f8b9e71f95937913adR1038).

inc/admin.php
```diff
@@ -83,6 +83,11 @@ function __construct() {
 				add_filter( 'display_post_states', array( $this, 'add_panels_post_state' ), 10, 2 );
 			}
 		}
+
+		// Compatibility with AMP plugin
+		if ( function_exists( 'amp_bootstrap_plugin' ) ) {
+			require_once plugin_dir_path( __FILE__ ) . '../compat/amp.php';
```
gregpriday, line 89: We should try keep all our compat includes in the same place. This is where we're adding another one - https://github.com/siteorigin/siteorigin-panels/blob/develop/siteorigin-panels.php#L197

What do you think about having a dedicated function in the `SiteOrigin_Panels` class that's run with `plugins_loaded` that does all our compat includes? Something like `init_compat`?

AlexGStapleton: Done. Do you want to move the SEO compatibility and Block Editor detect code (it's only there to load compatibility code) there also? If it's a yes for Yoast, I'll need to move `enqueue_seo_compat` to `siteorigin-panels.php`.

Issue #715, opened 2020-01-03
Block Editor: Prevent saving during Ajax


AlexGStapleton:
[Related thread](https://siteorigin.com/thread/page-bulder-on-gutenberg-seems-to-need-clicking-update-twice/#comment-348792)

Unlike the Classic Editor, there's currently nothing in place to prevent the Block Editor from saving when the SiteOirgin Layouts block is setting up widgets using Ajax. This can result in changes not saving when SiteOrigin Widgets Bundle is being used on a slow server or if the user is rapidly adding a large number of widgets. For a visual example of this issue, please give this user-submitted screencast a quick watch:

https://www.youtube.com/watch?v=VuhX245ie8k

As demonstrated, the easiest way to replicate this issue is to duplicate a large number of widgets and then quickly save.

PR #776, opened 2020-07-08
Layout Block: Prevent Block Editor Saving until Json loaded

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/715

This PR will prevent the user from being able to save the page until Page Builder has loaded the new layouts json and stored it.

### DIFF
--- a/compat/js/siteorigin-panels-layout-block.js
+++ b/compat/js/siteorigin-panels-layout-block.js
@@ -278,6 +278,7 @@ registerBlockType('siteorigin-panels/layout-block', {
     var onLayoutBlockContentChange = function onLayoutBlockContentChange(newPanelsData) {
       if (!_.isEmpty(newPanelsData.widgets)) {
         // Send panelsData to server for sanitization.
+        wp.data.dispatch('core/editor').lockPostSaving();
         jQuery.post(panelsOptions.ajaxurl, {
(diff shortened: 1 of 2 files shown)


compat/js/siteorigin-panels-layout-block.js
```diff
@@ -278,6 +278,7 @@ registerBlockType('siteorigin-panels/layout-block', {
     var onLayoutBlockContentChange = function onLayoutBlockContentChange(newPanelsData) {
       if (!_.isEmpty(newPanelsData.widgets)) {
         // Send panelsData to server for sanitization.
+        wp.data.dispatch('core/editor').lockPostSaving();
```
Misplon, line 281: Do you perhaps want to format this line and the line below with the same spacing as the changes to `siteorigin-panels-layout-block.jsx` ?

AlexGStapleton: This file is automatically generated by the build process, and it's removing the spacing inside of the parentheses. I can change its formatting, but it'll be replaced upon running the build process.

Misplon: Thanks for letting me know, yeah, that's no problem then.

Issue #718, opened 2020-01-11
Mobile: Difficult to reach widget action links


Misplon:
It's quite difficult to get to the widget action (Edit - Duplicate - Delete) links on mobile. If you very carefully tap on the widget description you can view the links but tapping almost anywhere else just opens the widget. Should we perhaps make the action links always visible on mobile?

**Screencast:** 
[RPReplay_Final1578753901.mp4.zip](https://github.com/siteorigin/siteorigin-panels/files/4049432/RPReplay_Final1578753901.mp4.zip)

Misplon:
@AlexGStapleton What do you reckon we do here? Perhaps make the action links always visible for mobile?

AlexGStapleton:
That's definitely the way to go. 

https://i.imgur.com/7d92dbB.png

That looks okay. Hm. What are your thoughts on hiding the description and moving the action links there? Something like this:

https://i.imgur.com/5KTbaZL.png

Or with the actions centered so they're easier to click:

https://i.imgur.com/KqKT57I.png

Misplon:
The last screenshot is a good usability idea but looks a little misaligned. Should we center the widget title?

AlexGStapleton:
That would look like this:
https://i.imgur.com/KjTjtrG.png

We could also set the actions to space between:
https://i.imgur.com/HZVwdos.png

Space between wIth centered title:
https://i.imgur.com/URmFX1n.png

How does the vertical alignment look in those screenshots? Do you want to move the actions up/down?
I feel like https://i.imgur.com/URmFX1n.png is the way to go as it's spread out more.

I've made the changes locally and will commit them once we've confirmed if we're going to center the title/alignment.

Misplon:
This is the one I think https://imgur.com/HZVwdos.

PR #777, opened 2020-07-08
Admin: Show Widget Action links on mobile devices

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/718.

The description is hidden to achieve this.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -494,6 +494,18 @@
 										}
 									}
 
+									@media (max-width:782px) {
+										position: static;
+										opacity: 1;
+										padding-left: 0;
+										display: flex;
+										justify-content: space-between;
+
+										a {
+											display: inline-block;
+										}
+									}
+
 								}
 							}
 
@@ -513,6 +525,10 @@
 								height: 16px;
 								overflow: hidden;
 								color: #777;
+
+								@media (max-width:782px) {
+									display: none;
+								}
 							}
 
 							.form{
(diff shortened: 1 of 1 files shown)


Misplon:
Do we perhaps want to rather use `is_mobile` for this?

AlexGStapleton:
We don't currently have a dedicated mobile-specific stylesheet for the admin interface. We can add one, but it seems a tad odd to do this for just this change so if we do this we should move all mobile CSS there. Do you want me to proceed?

I should mention that the breakpoint I'm using is the same value consistently used throughout the stylesheet.

Misplon:
That makes sense. Do we perhaps want to expand this change so that it applies to say an iPhone 10 in landscape?

Misplon:
The widget title is resizing in Chrome when you switch to landscape. I thought I had previously resolved this. Do you want to fix that in this PR if you can replicate?

IMG_0681

Misplon:
`-webkit-text-size-adjust: 100%;` is the solution.

AlexGStapleton:
I can't replicate that in Chrome. It may be a MacOS specific issue. Can you commit the suggested fix? I mean, I can but I can't check if it works before committing.

Misplon:
Thanks, sorted.

Do you want to take a look at the Add Widget search form and try to improve it for mobile?

Misplon:
We can do that in another PR though.

PR #778, opened 2020-07-09
Fix "undefined" Page Builder tab issue in WP 5.5

AlexGStapleton:
This PR resolves a few issues with the page builder interface in WP 5.5. 

Fix "undefined" Page Builder tab issue in WP 5.5
Add_New_Page_‹_SiteOrigin_—_WordPress

Fix "undefined" Page Builder tab issue in WP 5.5. For whatever reason, the span inside of the h2 was removed in WordPress 5.5. Removing the span from the selector resolves this issue and results in no adverse effects in either version of WP.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -18,7 +18,9 @@
 
 		margin-top: 20px;
 
-		.hndle, .handlediv {
+		.postbox-header,
+		.hndle,
+		.handlediv {
 			display: none !important;
 		}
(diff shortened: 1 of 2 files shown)


Misplon:
@AlexGStapleton Can you also replicate the following?

AlexGStapleton:
That's been removed in the last commit.

Issue #158, opened 2016-02-10
Option to disable parallax on mobile


gregpriday:
We've heard from users getting issues with parallax on mobile. It would be useful to give them the option to disable it for mobile devices.

Wachu91:
+1

AlexGStapleton:
@Misplon Do you want this setting added to the general tab above/under Limit Parallax Motion, or on the layouts tab? If the latter, where about's do you want to add the setting?

How do this sound?
Setting: Disable Background Parallax On Mobile
Description: Disable row/widget background parallax when browser is smaller than the mobile width.

Misplon:
Let's add this setting below the Limit Parallax Motion setting. Good to go with the suggested label and description, thanks.

PR #780, opened 2020-07-10
Add "Disable Background Parallax On Mobile" Global Setting

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/158

Requires premium PR: https://github.com/siteorigin/siteorigin-premium/pull/403
The premium is required because it overrides the PB parallax with its own (modified) verison.

### DIFF
--- a/inc/settings.php
+++ b/inc/settings.php
@@ -122,6 +122,7 @@ class SiteOrigin_Panels_Settings {
 		$defaults['admin-post-state']       = true;
 		$defaults['admin-widget-count']     = false;
 		$defaults['parallax-motion']        = '';
+		$defaults['parallax-mobile']        = false;
 		$defaults['sidebars-emulator']      = true;
 		$defaults['layout-block-default-mode'] = 'preview';
 
--- a/inc/settings.php
+++ b/inc/settings.php
@@ -286,8 +286,8 @@ class SiteOrigin_Panels_Settings {
 
 		$fields['general']['fields']['parallax-mobile'] = array(
 			'type'        => 'checkbox',
-			'label'       => __( 'Disable Background Parallax On Mobile', 'siteorigin-panels' ),
-			'description' => __( 'Disable row/widget background parallax when browser is smaller than the mobile width.', 'siteorigin-panels' ),
+			'label'       => __( 'Disable Parallax On Mobile', 'siteorigin-panels' ),
+			'description' => __( 'Disable row/widget background parallax when the browser is smaller than the mobile width.', 'siteorigin-panels' ),
 		);
 
 		$fields['general']['fields']['sidebars-emulator'] = array(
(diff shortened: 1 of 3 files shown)


inc/settings.php
```diff
@@ -284,6 +284,12 @@ function settings_fields( $fields ) {
 			'description' => __( 'How many pixels of scrolling result in a single pixel of parallax motion. 0 means automatic. Lower values give more noticeable effect.', 'siteorigin-panels' ),
 		);
 
+		$fields['general']['fields']['parallax-mobile'] = array(
+			'type'        => 'checkbox',
+			'label'       => __( 'Disable Background Parallax On Mobile', 'siteorigin-panels' ),
+			'description' => __( 'Disable row/widget background parallax when browser is smaller than the mobile width.', 'siteorigin-panels' ),
```
Misplon, line 290: Let's perhaps change this to: `Disable row/widget background parallax when the browser is smaller than the mobile width.`

It's quite long already so adding `the` won't hurt I don't think.

inc/settings.php
```diff
@@ -284,6 +284,12 @@ function settings_fields( $fields ) {
 			'description' => __( 'How many pixels of scrolling result in a single pixel of parallax motion. 0 means automatic. Lower values give more noticeable effect.', 'siteorigin-panels' ),
 		);
 
+		$fields['general']['fields']['parallax-mobile'] = array(
+			'type'        => 'checkbox',
+			'label'       => __( 'Disable Background Parallax On Mobile', 'siteorigin-panels' ),
```
Misplon, line 289: The setting above is "Limit Parallax Motion". What do you think about changing this label to "Disable Parallax On Mobile" ?

inc/styles.php
```diff
@@ -54,14 +54,18 @@ static function register_scripts() {
...
 		wp_register_script(
 			'siteorigin-parallax',
 			siteorigin_panels_url( 'js/siteorigin-parallax' . SITEORIGIN_PANELS_JS_SUFFIX . '.js' ),
 			array( 'jquery' ),
 			SITEORIGIN_PANELS_VERSION
 		);
-		wp_localize_script( 'siteorigin-panels-front-styles', 'panelsStyles', array(
-			'fullContainer' => apply_filters( 'siteorigin_panels_full_width_container', siteorigin_panels_setting( 'full-width-container' ) )
+		wp_localize_script( 'siteorigin-parallax', 'parallaxStyles', array(
+			'parallax-mobile' => ! empty( siteorigin_panels_setting( 'parallax-mobile' ) ) ? siteorigin_panels_setting( 'parallax-mobile' ) : false,
```
gregpriday, line 67: Can we use `?:` with the version of PHP we're supporting yet?

AlexGStapleton: Yes. Just submitted a commit using that instead..

Issue #727, opened 2020-02-14
Parallax: Update to use transform


Misplon:
Considering improving performance and usability by switching parallax from background-position to transform.

https://developers.google.com/web/updates/2016/12/performant-parallaxing

AlexGStapleton:
Do you want me to try and minimize the number of additional divs used, or are the 3 divs Google uses okay?

The library Google came up with (which is linked at the very end of the article you posted) looks to move the parallax to a separate div and reduces the overall number of elements.
([demo here](https://googlechromelabs.github.io/ui-element-samples/parallax/) - [code here](https://github.com/GoogleChromeLabs/ui-element-samples/tree/gh-pages/parallax)).  The library also handles the calculations which prevents us from having to do that. 

Both the article and the library they came up with are from 2016 so it might be worth checking if there's a more modern library/technique. Do you agree? If so, I'll take a look around.

Misplon:
Look around, please.

AlexGStapleton:
It seems most parallax libraries ignore CSS backgrounds in favor of HTML images. I can definitely understand their reasoning for this (main reason likely being actual control over the sizing of the image without fighting `background-size: cover;`). The only libraries I could find that supported CSS backgrounds were either dead in the water (no updates for a number of years with a tonne of pending pr and issues), used partial implementations, or used `background-position`.

It might be worth considering outputting the image directly rather than using a background. If you would (understandably) rather not take this path, we'll need to implement the versions highlighted in the above Google links.

if you want to consider the HTML Image route, here are some popular options:

https://github.com/geosigno/simpleParallax.js/
Focused on HTML images. Scaling options, but nothing mobile specific. Actively supported.

https://github.com/dixonandmoe/rellax
Has mobile-specific scaling. Actively supported.

Misplon:
Sounds like using the Google option is the best overall choice for us at the moment.

Issue #721, opened 2020-01-16
Add Filter for Row/Cell/Widget Styles Field Data


AlexGStapleton:
There's currently no way to filter row/cell/widget styles based on inputted data. This prevents developers from conditionally hiding styles based on the previously input data which can result in a cluttered interface due to a mixture of legacy and new fields.

PR #782, opened 2020-07-14
Parallax rework

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/727 https://github.com/siteorigin/siteorigin-panels/issues/721

This PR replaces the old parallax with a new img based parallax that uses [simpleParallax.js](https://simpleparallax.com/). A library was used for better overall compatibility and due to the added support it offers. Other changes:

- Add `siteorigin_panels_inside_CONTEXT_before/after`
- CSS added to both flex and legacy stylesheets to prevent display issues
- Removed parallax (original) due to it being made redundant (it's not really a thing with HTML images)
- Graceful fallback in IE 11 (no parallax, but image still displays correctly)
- Adds `siteorigin_panels_general_current_styles` and `siteorigin_panels_general_current_styles_TYPE` (type being the current styles type). This resolves https://github.com/siteorigin/siteorigin-panels/issues/721

Things left to do: 
- [x] Implement a filter that handles style migration to handle Parallax (original) to Parallax conversion.
- [x] Decide on new parallax settings
- [x] Implement new parallax settings
- [x] Add support for https://github.com/siteorigin/siteorigin-panels/pull/780
- [x] Update SiteOrigin Premium Parallax addon to use new parallax (separate PR https://github.com/siteorigin/siteorigin-premium/pull/432).

We should remove Limit Parallax Motion* and introduce the following settings:

**Parallax Delay** (default: 0.4)
The delay before the parallax finishes after the user stops scrolling.

**Parallax Scale** (default: 1.4)
How much the image is scaled

For more information on the above settings, please refer to https://simpleparallax.com/#settings


AlexGStapleton:
@Misplon * Limit Parallax Motion should be removed because how the parallax is handled now is completely different so the previous option doesn't translate. We can try and automatically convert it to a different setting/value but there's no telling what the user previously set it to.

Thoughts?

inc/renderer.php
```diff
@@ -539,6 +539,10 @@ function the_widget( $widget_info, $instance, $grid_index, $cell_index, $widget_
 			$args['after_widget'] = '</div>' . $args['after_widget'];
 		}
 
+		// This allows other themes and plugins to add html inside of the cell before and after its output.
```
Misplon, line 542: Perhaps capitalize html here "HTML".

css/front-legacy.less
```diff
@@ -3,6 +3,26 @@
...
+	.so-parallax {
+		position: relative;
+
+		& > div:not(.simpleParallax) {
+			position: relative;
+			z-index: 1;
+		}
+
+		.simpleParallax,
+		img[data-siteorigin-parallax] {
```
Misplon, line 16: Would you like to order selectors alphabetically for new rules as we do in themes?

Misplon:
@AlexGStapleton Can we write a formula to convert the limitation setting value into a matching scale value? If not, we should consider an admin notice to let users know about the change and where to adjust the new Scale setting.

Misplon:
Functionality looking good.

AlexGStapleton:
> @AlexGStapleton Can we write a formula to convert the limitation setting value into a matching scale value? If not, we should consider an admin notice to let users know about the change and where to adjust the new Scale setting.

We could, yes.

AlexGStapleton:
What are your thoughts on the suggested settings? @Misplon The suggested defaults are the simpleParallax.js defaults but we could set them to anything else so if you have a preference let me know.

Misplon:
Thanks, the defaults seem fine.

Misplon:
If you'd like to change them, let me know or make the change and I'll test.

AlexGStapleton:
To confirm, you're good with the suggested setting names and setting descriptions?

Misplon:
Names are good. Let's perhaps try the following as descriptions:

"The delay before the parallax effect finishes after the user stops scrolling."

"How much the image is scaled. The higher the scale is set, the more visible the parallax effect will be."

AlexGStapleton:
> "How much the image is scaled. The higher the scale is set, the more visible the parallax effect will be."

That's much better. I do however think we need to explicitly mention that this will result in the image scaling.

Misplon:
Perhaps:

"How much the image is scaled. The higher the scale is set, the more visible the parallax effect will be. Increasing the scale will result in a loss of image quality."

Misplon:
Thanks, mate. Please, try the One Page Layout in Corp and scroll down to the Fixed Background row. I made that row background Parallax, it was then hidden, I changed it back to Fixed and can't seem to get the row background to display. Can you replicate that issue?

AlexGStapleton:
Let me know if it's working as expected with the last commit.

Misplon:
Thanks for adding fallback support, that's helpful for prebuilt layouts. 
I should have mentioned that I moved the background local so fallback wasn't the issue.
Using a fallback image I see the following notice:

`Notice: Trying to access array offset on value of type bool in /siteorigin/wp-content/plugins/siteorigin-panels/inc/styles.php on line 440`

Using the image from the Media Library with the build dev task running, the row is broken:

Screenshot 2020-07-28 at 19 27 10

AlexGStapleton:
I'm not getting that error message, but I can see a potential reason why it may happen so I've accounted for it in the last commit.

Regarding the background showing like that, that looks like the parallax CSS isn't outputting, or maybe even a class isn't being added. Can you show me a screenshot of the row with .simpleParallax container selected?

It should look like this:



Purely for reference, [here's a copy of the build](https://drive.google.com/uc?id=1qYyR_vn3nmkyvvX5BVFC23g2ihkwOklS) I'm using.

Misplon:
@AlexGStapleton Thanks, mate. Please, can you update the original PR description and move the necessary items from the To  Do to Done list to make reading easier for when Greg reviews.

Misplon:
@AlexGStapleton Does the Limit parallax motion setting still need to be removed?

AlexGStapleton:
> If the row was previously using Parallax (original size), this branch will set the selection to Tiled.

I was able to replicate this, but upon updating the branch this issue appears to have resolved itself.

Misplon:
@AlexGStapleton using the below layout are you able to see this issue where changing from develop to this branch will change the image background position?

original

new

[home.json.zip](https://github.com/siteorigin/siteorigin-panels/files/5334469/home.json.zip)

AlexGStapleton:
What you're seeing in the previous version is the fallback image rather than the parallax. Prior to this PR if the parallax wasn't able to parallax, the image would be vertically centered using background-position: cover; and no parallax would occur. If a user brings this up, we should encourage them to switch the **Background Image Display** setting to **Cover** as that's what they were using prior to this update.

Regarding the image being cut off at the bottom with the new parallax, if the area the image is applied too isn't large enough for the size of the image, simpleParallax will simply cut the bottom of the image off. The previous parallax would parallax more when this happened (or not parallax at all) which looked quite odd if you had several parallaxes as each one could be a different speed. If the user wants more of the image to show, we should encourage them to either increase the size of the area that the parallax is applied to using padding/height, or suggest decreasing the height of the image.

Misplon:
Thanks for the above explanation.

Misplon:
@AlexGStapleton Are we ready here? Do you want to tick the last checkbox above? Good work.

AlexGStapleton:
@Misplon Yes, this PR is ready for final review.

> Do you want to tick the last checkbox above? 
Just marked it.

Misplon:
Awesome!

Misplon (approved):

gregpriday (approved):
I imagine this is a big change that might well cause some problems. I've checked over this code a few dozen times and can't find anything wrong. We just need to make sure we test it as much as possible.

AlexGStapleton:
I've talked to the rest of the team about how to handle the upgrade process and we've decided that we have to maintain the previous parallax. We talked about two possible ways of doing this:

Option 1: Filters

- After update/install, check if legacy parallax settings are present.
- If set, set an option called `siteorigin_legacy_parallax` (this is required because those legacy settings will be removed if the user saves their PB settings) and that option is a serialized array of their legacy settings at the time of the update.
- When outputting the parallax code and changes, we check if siteorigin_legacy_parallax" exists and act accordingly.
- Add a filter that allows you to change parallax regardless of option value.
- Add filer to change legacy parallax settings (only new parallax will have dedicated settings)

Option 2: Dedicated settings (no PHP required by user)

- Add either a select (name undecided, maybe just Parallax Type, or checkbox (Use Legacy Parallax)
- After update/install, check if legacy parallax settings are present.
- If set, set a page builder setting to `legacy` (or `true`)
- If not set, set page builer setting to `new` (or `false`)
- Conditionally show relevant parallax settings based on parallax value.
- When outputting the parallax code and changes, we check the page builder setting and act accordingly.

We've decided on the second option. While the first option would be quicker, it wouldn't be very user friendly and would be tricky to adjust.

AlexGStapleton:
The last commit does:

- Reintroducee legacy parallax settings.
- Add Parallax Type setting with conditional default for parallax_type setting.
- Add legacy parallax JS file and renamed it to `siteorigin-legacy-parallax.js.`
- Add helper method to detect parallax background `SiteOrigin_Panels_Styles::is_background_parallax().
- Prevent New Parallax from running if `window.parallaxStyles` isn't set.

It wasn't my intention for this to be done in a single commit, but due to the scope of this change, it was proving tricky to split them up into logical separate commits that still functioned.

AlexGStapleton:
I wasn't sure what description to give the Parallax Type setting. It's currently:

>  Which Parallax library will be used. The New is the recommended Parallax Type as it uses modern techniques. 

Open to suggestions.

frafor1988:
Hi Alex, just as a suggestion. I ended up here because a client of mine using SOPB said the parallax in his site was slowing down scrolling. So from a user perspective the description could be like:

`Which Parallax library will be used. The New is the recommended Parallax Type as it provides better performance on modern browsers.`

Btw thanks for all your work.

AlexGStapleton:
Hi @frafor1988,

Thanks for commenting! :)
I've forwarded your suggestion to the rest of the team and we're currently talking about it. We may actually change the name of the parallax types as "new" isn't very descriptive. Modern has been suggested.

I've prepared you a build so you can try the current working version of this branch if you would like to. Here's the download link:

[siteorigin-panels.new-parallax.zip](https://github.com/siteorigin/siteorigin-panels/files/5883336/siteorigin-panels.new-parallax.zip)

I don't recommend using this on a production site, but if you're able to tell it on a development/staging version of the site that's facing poor performance and provide feedback that would be much appreciated. You can install it by either dropping the contents into an existing siteorigin-panels directory or by navigating to **Plugins > Installed Plugins** and deactivate Page Builder by SiteOrigin, then delete it. Scroll to the top and click **Add New**, then **Upload Plugin**. Upload siteorigin-panels.new-parallax.zip and when prompted, activate Page Builder by SIteOrigin.

After adding the updated version, please navigate to **Settings > Page Builder** and open the **General** tab. Go down to the **Parallax Type** setting and set the setting to **New**. Adjust the newly visible Parallax settings as desired.

Kind regards,
Alex

gregpriday (approved):

PR #786, opened 2020-07-18
Improve Block Editor SEO Compatibility Support

AlexGStapleton:
This PR improves the SEO compatibility code for the SiteOrigin Layout Block. It allows for multiple SiteOrigin Layout Blocks to pass changes to SEO Plugins and also resolves a content extraction issue for the SiteOrigin Layout Block that prevented certain widgets from being completely compatible. An example of such a widget is the SiteOrigin Block Editor widget where the Destination Link setting wasn't being picked up correctly.

This is quite a large change with a lot of moving parts so thorough testing is required. This PR must also be tested using either `build:dev` or `build:release` as editor JS files were modified.

### DIFF
--- a/compat/js/siteorigin-panels-layout-block.jsx
+++ b/compat/js/siteorigin-panels-layout-block.jsx
@@ -118,6 +118,12 @@ class SiteOriginPanelsLayoutBlock extends Component {
 		} );
 		
 		jQuery( document ).trigger( 'panels_setup', this.builderView );
+
+		if ( typeof window.soPanelsBuilderView == 'undefined' ) {
+			window.soPanelsBuilderView = [];
+		}
+			window.soPanelsBuilderView.push( this.builderView );
+		}
 		
 		this.panelsInitialized = true;
 	}
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -731,8 +731,15 @@ module.exports = Backbone.View.extend( {
 	handleContentChange: function () {
 
 		// Make sure we actually need to copy content.
-		if ( panelsOptions.copy_content && this.attachedToEditor && this.$el.is( ':visible' ) ) {
-
+		if ( panelsOptions.copy_content	&&
+			(
+				typeof wp.blocks !== 'undefined' ||
+				(
+					this.attachedToEditor &&
+					this.$el.is( ':visible' )
+				)
+			)
+		) {
 			var panelsData = this.model.getPanelsData();
 			if ( !_.isEmpty( panelsData.widgets ) ) {
 				// We're going to create a copy of page builder content into the post content
--- a/js/siteorigin-panels/view/builder.js
+++ b/js/siteorigin-panels/view/builder.js
@@ -731,15 +731,7 @@ module.exports = Backbone.View.extend( {
 	handleContentChange: function () {
 
 		// Make sure we actually need to copy content.
-		if ( panelsOptions.copy_content	&&
-			(
-				typeof wp.blocks !== 'undefined' ||
-				(
-					this.attachedToEditor &&
-					this.$el.is( ':visible' )
-				)
-			)
-		) {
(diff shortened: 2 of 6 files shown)


Misplon:
@AlexGStapleton Not sure if I was meant to but I imported https://github.com/siteorigin/siteorigin-panels/pull/794 into my local version of this (improve-block-editor-seo-support) branch. Can you please, check. When you test this branch on the Corp home page (the bundled version is fine so you can test locally). Is Yoast in the Block Editor seeing outbound links? It isn't for me. I'm wondering if that's just a caching issue.

AlexGStapleton:
Did you run `build:dev` prior to testing? There's changes made to a core builder file so the dev task must be run (including when making a release build) to compile the updated JS. For reference, the specific JS is [here](https://github.com/siteorigin/siteorigin-panels/pull/786/files#diff-9900a73a100c9dc614ce2adb1d10a6b2R734-R742).

I just ran a test and when https://github.com/siteorigin/siteorigin-panels/pull/794 is merged this PR continues to work as expected.

Misplon:
I did yeah, I'll try again tomorrow. Thanks.

Misplon:
I can see the updated JS on the page so I'm using the changes. I'm still not seeing the outbound link contained in the Hero in the Corp home page layout.

_Outbound links: No outbound links appear in this page. Add some!_

Using the Corp home page JSON layout in the Block Editor with this branch, are your outbound links accounted for?

Misplon:
I've uploaded a build to a live site, the outbound link issue isn't present. Can you perhaps update to the 5.5 beta (5.5-beta3-48630) and confirm that this branch is working normally there? I ask because that's a possible difference between my test site and localhost.

AlexGStapleton:
I can confirm outbound links are being picked up for me while using `5.5-beta4-48649`.

Screenshot_2020-07-28 Edit Page ‹ blueprint — WordPress

Misplon:
Thanks for checking.

compat/js/siteorigin-panels-layout-block.jsx
```diff
@@ -118,6 +118,12 @@ class SiteOriginPanelsLayoutBlock extends Component {
 		} );
 		
 		jQuery( document ).trigger( 'panels_setup', this.builderView );
+
+		if ( typeof window.soPanelsBuilderView == 'undefined' ) {
+			window.soPanelsBuilderView = [];
+		}
```
gregpriday, line 124: It looks like something strange is going on with this line. Is this closing `}` meant to be here?

js/siteorigin-panels/view/builder.js
```diff
@@ -731,8 +731,15 @@ module.exports = Backbone.View.extend( {
 	handleContentChange: function () {
 
 		// Make sure we actually need to copy content.
-		if ( panelsOptions.copy_content && this.attachedToEditor && this.$el.is( ':visible' ) ) {
-
+		if ( panelsOptions.copy_content	&&
```
gregpriday, line 734: I agree with [this answer](https://softwareengineering.stackexchange.com/a/3451) on StackExchange that if an if statement gets big, it becomes unreadable. We should probably refactor it into a well named function.

compat/js/siteorigin-panels-layout-block.jsx
AlexGStapleton: Yes. If `window.soPanelsBuilderView` isn't initialized, this creates it as an empty array.

js/siteorigin-panels/view/builder.js
AlexGStapleton: How does `isContentCopyable` or `isCopyContentEnabled` sound? If neither, do you have any suggestions?

Hm. Actually, we could go further and split this up into three functions - `isCopyContentEnabled`, `isClassicEditor`, `isBlockEditor`.

gregpriday: I don't have any strict preferences here, so which ever you feel is neater works for me 🚀

gregpriday (approved):

PR #787, opened 2020-07-19
Remove version from core JS files urls

AlexGStapleton:
Requires https://github.com/siteorigin/plugin-build/pull/16


Misplon:
I've tested a build and am not seeing any issues at this time.

AlexGStapleton:
I'm not actually sure how to do that so @Misplon will need to do that.

Issue #791, opened 2020-07-22
Since update to 2.11.1 we are receiving "Notice: Undefined index: panels_info"


korridor:
The error occurs at:   
`/wp-content/plugins/siteorigin-panels/inc/widgets/post-loop.php` in `SiteOrigin_Panels_Widgets_PostLoop::widget` at line 153

AlexGStapleton:
HI Korridor,

Thank you for reporting this warning. I've identified why this is happening and I've come up with a fix. https://github.com/siteorigin/siteorigin-panels/pull/792

Kind regards,
Alex

korridor:
Thanks for the quick response and fix.

Best regards,
Korridor

Misplon:
Hi @korridor 

Thanks for logging :) I'm going to close this issue, if there are any issues after the next release, please, let us know. We're working on the last PR's for the release this week. If you'd like a beta copy to use now, please, let me know. Thanks again.

PR #792, opened 2020-07-22
Post Loop: Resolve notice when used outside of Page Builer

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/791

### DIFF
--- a/inc/widgets/post-loop.php
+++ b/inc/widgets/post-loop.php
@@ -75,6 +75,7 @@ class SiteOrigin_Panels_Widgets_PostLoop extends WP_Widget {
 	 * @return array
 	 */
 	function update( $new, $old ){
+		$new['pagination_id'] = md5( json_encode( $new ) );
 		if( class_exists( 'SiteOrigin_Widget' ) && class_exists( 'SiteOrigin_Widget_Field_Posts' ) ) {
 			$helper = $this->get_helper_widget( $this->get_loop_templates() );
 			return $helper->update( $new, $old );
@@ -150,10 +151,18 @@ class SiteOrigin_Panels_Widgets_PostLoop extends WP_Widget {
 				if(!empty($matches[1])) $query_args['paged'] = intval($matches[1]);
 				else $query_args['paged'] = 1;
 			} else {
-				self::$current_pagination_id = substr( $instance['_sow_form_id'], 0, 7 );
-				if ( isset( $_GET[ 'page-' . self::$current_pagination_id ] ) && is_numeric( $_GET[ 'page-' . self::$current_pagination_id ] ) ) {
-					$query_args['paged'] = $_GET[ 'page-' . self::$current_pagination_id ];
-				} else {
+				if ( isset( $instance['pagination_id'] ) ) {
+					self::$current_pagination_id = $instance['pagination_id'];
+
+					if (
+						isset( $_GET[ 'page-' . self::$current_pagination_id ] ) &&
+						is_numeric( $_GET[ 'page-' . self::$current_pagination_id ] )
+					) {
+						$query_args['paged'] = $_GET[ 'page-' . self::$current_pagination_id ];
+					}
+				}
+
+				if ( ! isset( $query_args['paged'] ) ) {
 					$query_args['paged'] = 1;
 				}
 			}
--- a/inc/widgets/post-loop.php
+++ b/inc/widgets/post-loop.php
@@ -150,7 +150,7 @@ class SiteOrigin_Panels_Widgets_PostLoop extends WP_Widget {
 				if(!empty($matches[1])) $query_args['paged'] = intval($matches[1]);
 				else $query_args['paged'] = 1;
 			} else {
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. This PR resolves #791 but it appears to break https://github.com/siteorigin/siteorigin-panels/pull/781, at least from what I can see in my local testing. I tested using a custom loop using the pagination suggested in the PR. I inserted two Post Loop widgets in the Classic Editor on the same page and selected a different taxonomy in each loop. Clicking to page 2 seems to impact both loops.

AlexGStapleton:
The last commit allows the previously provided PHP to work.

Misplon:
Thanks. This only works if one of the two loops in use doesn't have more pages than each other. Using the new theme unit test data, the classic category produces four pages and the block category produces 2 pages. Each loop hangs onto its page as long as I don't go over the lowest max. So if I select page 2 for the classic loop, the block loop stays on page 1. But I select page 4 for the classic loop, the block loop goes to not found.

Home_-_SiteOrigin

AlexGStapleton:
That appears to be an issue with Corp. I can't replicate this while using a different theme and rolling back to PB `2.10.17` (before the initial PR) still results in the 404 for the second query. Are you able to replicate this while using a different theme?

Misplon:
In North, I'm getting the following:

```Notice: Undefined index: _sow_form_id in /Users/misplon/Sites/siteorigin/wp-content/plugins/siteorigin-panels/inc/widgets/post-loop.php on line 153

Notice: Undefined index: _sow_form_id in /Users/misplon/Sites/siteorigin/wp-content/plugins/siteorigin-panels/inc/widgets/post-loop.php on line 153```

Misplon:
Ok, those notices appear after changing themes. It appears as though the Post Loop widget has changed to a template the new theme has but the widget is holding onto the previous template value.

Misplon:
Page_Builder_-_SiteOrigin

The same issue persists in North.

To confirm, loop one on the page has four pages. Loop two on the page has two pages. If I select page 2 in loop one, loop two will remain on page 1. But if I select a page in loop one that loop two doesn't have, loop two goes 404. Can you confirm this is the test you've run?

AlexGStapleton:
I can confirm that's the test I ran, and I can't replicate this issue while not using Corp.

I can replicate this issue while not using the snippet provided in #781 and while using an older version of PB. Are you able to replicate this issue if you're not using the snippet provided in #781? If so, are you able to replicate this while using PB `2.10.17`?

AlexGStapleton:
> In North, I'm getting the following:

It turns out that id isn't reliable either. The last commit will generate a random number when the post loop widget is saved and that will be used for the pagination. `rand()` isn't truly unique, but it's good enough for our usage.

Misplon:
Does the custom pagination need to be adjusted? Is the following still good?

```	<?php echo paginate_links( array(
		'format'       => '?page-' . SiteOrigin_Panels_Widgets_PostLoop::get_current_pagination_id() . '=%#%',
		'prev_next'    => false,
	) );```

AlexGStapleton:
Yes, that code is still fine. Please note that you must save the Post Loop widget for the ID used by the pagination to be generated. If you don't resave the widget, the provided code won't work as expected. (it'll come up as `?page-=2`).

Regarding the 404 issue you're facing, that's not going to be resolved by the last commit. My last question still remains:
> Are you able to replicate this issue if you're not using the snippet provided in #781? If so, are you able to replicate this while using PB 2.10.17?

Misplon:
The 404 issue persisted before this PR, yes.

Misplon:
At the moment, if I change the page in the normal loop it's impacting the custom pagination loop but not the other way around. Here is a screencast showing my testing setup.

[post-loop-custom.mov.zip](https://github.com/siteorigin/siteorigin-panels/files/5011986/post-loop-custom.mov.zip)

Misplon:
Unless I've misunderstood the testing required here, are both loops meant to be using custom pagination?

Misplon:
Ok, both using custom loops works, my bad.

AlexGStapleton:
Thank you for the video. That's a tricky situation. I'm unsure how we can really get around that as that's happening as we're still looking for the base pagination. Hm. We could add a filter to selectively disable the standard pagination (basically, bypass the other checks) on a widget by widget basis. Yeah, that feels like the way to go. How does `siteorigin_panels_post_loop_custom_pagination` or `siteorigin_panels_post_loop_use_custom_pagination` sound? It would be a boolean check.

Misplon:
`siteorigin_panels_post_loop_custom_pagination` sounds good. Thanks.

AlexGStapleton:
The last commit adds `siteorigin_panels_post_loop_custom_pagination`. To enable it please add `add_filter( 'siteorigin_panels_post_loop_custom_pagination', '__return_true' );` to your test site.

Misplon:
Looks good. We ready to move this onto review?

Issue #703, opened 2019-10-09
Android Chrome: Unable to open widgets


AlexGStapleton:
A user reported that they aren't able to open widgets while using Chrome on Android, and I've been able to replicate this issue. It looks like the "drag and drop" code for the widgets kicks in as the widget "jumps" briefly after selection.

This cannot be replicated while using the Chrome DevTools Device Mode.

Misplon:
Are there any console errors to report?

AlexGStapleton:
Chrome has something called [Remote Debugging](https://developers.google.com/web/tools/chrome-devtools/remote-debugging). Unfortunately, I'm not getting any error messages. :(

Misplon:
Thanks

@braamgenis Are you able to replicate?

Misplon:
@AlexGStapleton If you could try to take another briefish look, as discussed on Slack, that would be great. Thanks.

Misplon:
@AlexGStapleton I forget if we spoke on Slack after this message above. Did we?

AlexGStapleton:
I can't seem to find anything related to this, but I'll take a look at this very shortly.

Misplon:
Thanks

AlexGStapleton:
Unfortunately, I'm not seeing any potential causes. 😞

Misplon:
Cool, thanks for checking again. Perhaps we'll ask Andrew Freeborn to take a look when there is time.

PR #793, opened 2020-07-24
Android Chrome: Improve edit widget touch detection

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/703

This PR resolves an issue affecting certain Android devices where the edit widget prompt wouldn't be opened when they touched the widget. Instead, the widget would go into the draggable mode. This PR allows for both the edit widget prompt to open and the widgets to still be draggable.


AlexGStapleton:
I was able to replicate this issue on a borrowed Samsung Tab T6 (which was released at the end of last year IIRC). It's possible this issue is specific to Samsung devices. I don't own any Samsung devices though so this is only a theory.

Issue #795, opened 2020-07-31
SO Layout Block: Investigate automatic wp_targeted_link_rel on links


AlexGStapleton:
The Block Editor will automatically update all links with a `target="_blank"` if the link doesn't already account for [this issue](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/). This, in turn, will break the SiteOrigin Layout Block as the saved content is different to the generated `save` content. 

This happened in the SiteOrigin Editor widget and we were able to account for it here: https://github.com/siteorigin/so-widgets-bundle/pull/1080 That sort of fix, however, isn't something we're able to automatically do for third party widgets. We should explore options to see if we can reliably apply `wp_targeted_link_rel` to the general Layout Block data without it negatively affecting the layout.

PR #797, opened 2020-08-01
Run wp_targeted_link_rel on the generated preview content

AlexGStapleton:
Resolves https://github.com/siteorigin/siteorigin-panels/issues/795

This commit prevents an issue where the SO Layout Block may break due to saved data being different.

To test this issue you'll need to have a widget that adds a link with `target="_blank"` set and doesn't set both `noopener` and `noreferrer` (related https://github.com/siteorigin/so-widgets-bundle/pull/1080). Add the widget, set the widget up to add the link and then save. Reload the page and the editor will SiteOrigin Layout Block will Break:

Load this PR and recover the block by clicking the three dots and then selecting `Attempt Block Recovery`. Duplicate a widget/row (this will trigger the block content to update itself) and then remove the duplicate. Save. Reload the editor and the Block will work as expected.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -1129,7 +1129,7 @@ class SiteOrigin_Panels_Admin {
 		unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
 
 		$GLOBALS[ 'SITEORIGIN_PANELS_PREVIEW_RENDER' ] = true;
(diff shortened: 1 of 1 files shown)


Misplon:
It was introduced in 5.1.0.

Misplon:
I've sent Alex a test layout via Slack that I'm having trouble with. I'll await his feedback and testing with that layout before merging.

Issue #798, opened 2020-08-04
Block Editor: Consider adding the page name as a prefix to the JSON export


Misplon:
Consider adding the page name as a prefix to the JSON export (Layouts > Import/Export > Download Layout).

PR #800, opened 2020-08-05
Layout Export: Fix post title detection in Block Editor

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/798

The title wasn't previously showing as the Block Editor title field is set up slightly differently.

### DIFF
--- a/js/siteorigin-panels/dialog/prebuilt.js
+++ b/js/siteorigin-panels/dialog/prebuilt.js
@@ -169,7 +169,7 @@ module.exports = panels.view.dialog.extend( {
 		c.find( '.so-export' ).submit( function ( e ) {
 			var $$ = $( this );
 			var panelsData = thisView.builder.model.getPanelsData();
-			var postName = $('input[name="post_title"]').val();
+			var postName = $( 'input[name="post_title"], .editor-post-title__input' ).val();
 			if ( ! postName ) {
 				postName = $('input[name="post_ID"]').val();
 			}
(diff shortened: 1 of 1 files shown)


Misplon:
I can't seem to get this branch to work as it should. My JSON exports are still labeled with the ID only. For example `23992.json`.

Misplon:
@AlexGStapleton Can you confirm that the JSON export file is being labelled correctly for you? Thanks.

AlexGStapleton:
I can confirm the export is working as expected for me. Did you run `build:dev`? Sorry, I should have mentioned this previously.

Misplon:
Thanks for confirming. I usually run the build script while testing. I'm not sure what the issue is, I'll re-test shortly and get back to you.

Misplon:
Thanks, this branch is now working. Should we not add a suffix with the SO Layout Block or something like that. Right now if you have two SO Layout Blocks and try download both your file names will be:

block-editor.json
block-editor (1).json

The (1) being issues by macOS in this case.

AlexGStapleton:
How does that look?

Misplon:
Is the page title being repeated for you? It seems to be for me:

block-editor-block-editor-0.json
block-editor-block-editor-1.json

AlexGStapleton:
Oh. I misread your last reply and had thought you wanted the file to be called "page-name-block-layout-1.json". I'll correct this shortly.

Misplon:
Looking good.

PR #801, opened 2020-08-05
Layout Block: Add Support for new Preview Options

AlexGStapleton:
This PR resolves an issue new [Block Editor Preview Options](https://make.wordpress.org/core/2020/08/04/new-editor-preview-options/) and layouts that contain Row Layouts set to something other than Standard. To replicate this issue please import the Consultant Layout into the Block Editor and change device preview. You'll notice that the length of the rows is larger than the device preview area. This PR triggers a row resize to resolve this issue.


Misplon:
The preview isn't perfect but it seems better than what we already have.

PR #804, opened 2020-08-18
Avoid potential JS i18n variable conflict

AlexGStapleton:
`__` is far too common to rely so this PR uses `wp.i18n.__` directly.


Misplon:
Thanks, mate. Are there additional changes in this PR that got in there accidentally?

AlexGStapleton:
The other code changes are the build scripts doing - all change to siteorigin-panels-layout-block.js is done by the build script.
I can remove them, but they'll just be re-added upon running `build:dev` or `build:release`.

Issue #321, opened 2016-12-14
Plugin incompatibility: Gravity Forms + Yoast SEO


d--j:
### What did you expect to happen?

The Page Builder should work with the popular plugins [Gravity Forms](http://www.gravityforms.com/) and [Yoast SEO](https://wordpress.org/plugins/wordpress-seo/).

If all three plugins are activated there should be no JavaScript errors and the Gravity Forms CSS should be enqueued.

### What happened instead?

When a Gravity Form is included on a page via the SiteOrigin Page Builder and Yoast SEO is active the jQuery library does not get enqueued – resulting in "jQuery is not defined" JavaScript errors.
Also the Gravity Forms CSS files do not get enqueued.

If you deactivate the Yoast SEO plugin the page loads without errors.

### How can we reproduce this behavior?

Install Gravity Forms, SiteOrigin Page Builder and Yoast SEO and add a page where the Page Builder is active and Gravity Form widget gets included.

Here is a test page:
https://yoast.test-umgebung.de/page-with-pagebuilder/

Here is a page where the form is included without the Page Builder (there the error does not occur):
https://yoast.test-umgebung.de/page-whitout-pagebuilder/

### Technical info
* WordPress version: 4.7
* Yoast SEO version: 4.0
* Gravity Forms: 2.1.1.19
* Page Builder by SiteOrigin: 2.4.20
* Theme: Twenty Sixteen or Twenty Fifteen
* Webserver: nginx, php 7, MySQL 5.7, Linux

Contact me privately to get WordPress backend access and/or SFTP access to the test installation.

See: Yoast/wordpress-seo#6263

Misplon:
Hi d--j 

Thanks for posting and for such a nicely formatted issue report. The team will take a look at this as soon as they're able.

From a non-technical perspective I wanted to quickly share with you my experience of this combination. You can see for example, Yoast, Gravity and PB in use [here](http://ctshipsuppliers.com/contact-us/). The workaround at that link is to use Gravity via a Text widget using the form shortcode. Perhaps you can try that and let me know if helps the problem, as a workaround that is.

d--j:
Hi @Misplon,

yes looks like that workaround works: https://yoast.test-umgebung.de/form-as-shortcode/

Misplon:
[...]

PR #806, opened 2020-08-24
Compatibility: Gravity Forms + Yoast fix

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/321

This PR will run [gform_disable_print_form_scripts](https://docs.gravityforms.com/gform_disable_print_form_scripts/) when the Gravity Forms widget is added to a Page Builder layout.

This PR adds `siteorigin_panels_widget_instance` ([docs PR](https://github.com/siteorigin/docs/pull/105))


Misplon:
Thanks. The issue seems to be persisting for me. Here is what I see:

AlexGStapleton:
The provided fix won't adjust the setting as that's overly complicated and potentially confusing ("why does this setting keep getting re-enabled?"). Instead, it overrides the setting on the fly using a filter.

Can you please [send me a copy of the form](https://docs.gravityforms.com/exporting-a-form-from-gravity-forms/) you're testing? To be clear, this issue is only visible when using the form, and not while editing it.

Misplon:
> The provided fix won't adjust the setting as that's overly complicated and potentially confusing ("why does this setting keep getting re-enabled?"). Instead, it overrides the setting on the fly using a filter.

Thanks, I'm not expecting this PR to adjust the setting, my screenshot is just showing everything I see.

> To be clear, this issue is only visible when using the form, and not while editing it.

Using the form I'm testing with, there isn't an error on the site front-end using the develop branch. Does the form need conditional fields?

When enabling the Disable script output setting, the JS error does resolve in the admin. Should we not also use the filter in the admin to prevent the error?

Misplon:
Test form: 
[gravityforms-export-2020-08-25.json.zip](https://github.com/siteorigin/siteorigin-panels/files/5122474/gravityforms-export-2020-08-25.json.zip)

AlexGStapleton:
I'm not getting the error message you're getting when editing the widget. I'm also having trouble replicating this error again. 

> Does the form need conditional fields?

Yes. [Here's the form](https://drive.google.com/uc?id=1f48BtNFq6KRvvSM7DClvcmIh8TQKS3TH) I was using when I initially created this PR. Mind you, I'm having trouble replicating it now. :/
I hate to say it, but it might be worth closing this and suggesting disabling "disable script output" if it happens. It feels safer.

Misplon:
Thanks, I've tried your form and I've tried adding a conditional field to my first test form, I'm now not able to recreate a console error in the admin or front-end.

compat/gravity-forms.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * Disable Gravity Disable Print Form Scirpts.
```
Misplon, line 3: Gravity Forms Disable Print Scripts?

Misplon (changes requested):
Description update/typo correction required, otherwise all good.

Misplon (approved):

Misplon:
@AlexGStapleton Can you take a look at the above conflict? Thanks, mate.

AlexGStapleton:
Done.

Issue #704, opened 2019-10-13
SiteOrigin Layout: Post Loop widget doesn't preview


AlexGStapleton:
The Post Loop widget doesn't display in the SiteOrigin Layout Block preview.

Here's a gif that showcases this:
![](https://i.imgur.com/8kqhC8Z.gif)

(please excuse the cursor position)

PR #813, opened 2020-09-16
SiteOrigin Layout: Allow Post Loop widget to be previewed

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/704


Misplon (approved):

Misplon:
@AlexGStapleton Is it expected that styling isn't applied? I suppose so since the styling is theme level styling?

AlexGStapleton:
Yes. If the theme loads its CSS in the Block Editor it'll appear correctly. Most themes, including ours, don't do this though.
This is the main reason why the Post Loop widget itself doesn't have a preview option.

Misplon:
Got you. Thanks.

gregpriday (approved):

Issue #811, opened 2020-09-16
Classic Editor: Yoast content analysis returning zero words with PB active but not used


Misplon:
Yoast is returning a zero word count. I haven't checked the Block Editor yet, just the Classic Editor.

gregpriday:
Did this start happening with one of our updates of one of their's?

Misplon:
Not sure. Will narrow it down shortly.

Misplon:
The issue is only present in Yoast 14.9. Rolling back resolves the issue.

Misplon:
The issue isn't consistent, the word count being returned is sometimes normal. We aren't yet sure why.

PR #814, opened 2020-09-16
SEO Compat: Ensure PB is enabled before running compatibility code

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/811

window.soPanelsBuilderView will always be set when using the Classic Editor so the compatibility code will always run. This is fine in the Block Editor, but it runs into issues with the Classic Editor. This PR is set up to only run the Compatability code when PB is enabled or the SiteOrigin Layout is present.

### DIFF
--- a/js/seo-compat.js
+++ b/js/seo-compat.js
@@ -17,7 +17,12 @@ jQuery(function($){
 	};
 
 	SiteOriginSeoCompat.prototype.contentModification = function( data ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
I've set a build to the user who reported this.

Misplon (approved):

AlexGStapleton:
User reported this PR resolved this issue for them.

gregpriday (approved):

PR #822, opened 2020-10-13
Add Layouts AJAX response filter

kallehauge:
We're in a situation where we would like to add a simple "Favorites" filter to the layouts response - so only posts with a certain meta data will be shown - but there is no way for us to hook in and do this.

One way to allow this is to add an option for early opt-in to take over the results (an `if` condition before the [prebuilt condition](https://github.com/siteorigin/siteorigin-panels/blob/develop/inc/admin-layouts.php#L198)?)

But we can also achieve the same results if we get a filter for the final response and I feel like this - in general - is just a good addition to developers. So for now this would be my "I hope they accept it quickly" suggestion as seen in the PR :relaxed:

Best regards

### DIFF
--- a/inc/admin-layouts.php
+++ b/inc/admin-layouts.php
@@ -308,7 +308,7 @@ class SiteOrigin_Panels_Admin_Layouts {
 			$return['title'] .= __( ' - Results For:', 'siteorigin-panels' ) . ' <em>' . esc_html( $search ) . '</em>';
 		}
 		
-		echo json_encode( $return );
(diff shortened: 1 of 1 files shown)


AlexGStapleton (approved):
Hi @kallehauge,

Thank you for submitting a PR.  I don't see any issues with the changes you've made and don't have any recommendations for how you could improve it. 👍 

Kind regards,
Alex

Misplon (approved):

Issue #816, opened 2020-09-22
Add setting to output Row/Cell Layout CSS in header


AlexGStapleton:
A user posted about improving their Cumulative Layout Shift score, and found good results by forcing [siteorigin-panels-layouts](https://github.com/siteorigin/siteorigin-panels/blob/721da1a82c5aeab6a9a8f3af018f05cb1ff8e679/inc/renderer.php#L593-L594) to output in the <head>. As a result of that, we should add a method of doing this via a setting. I'm suggesting a situation as there are situations where this change may cause a negative impact on the user's score.

We should add this setting to the Page Builder Layouts tab. The setting name and description should be:

**Output Page Builder Layout CSS in header**
Outputting the CSS in the header may improve Cumulative Layout Shift score

AlexGStapleton:
@Misplon I've started on this but I've realized we may need to change how this works. Instead of a setting that forces the CSS to output in the header, we should introduce a setting called:

 **Output Page Builder Layout CSS**
Depending on your setup, outputting panels CSS to either the header or footer may improve improve performance.

Values:

Automatic (current)
Header (or head)
Footer

This will allow the user to force the CSS to be added to the head (which may potentially reduce Cumulative Layout Shift), or footer (which may also reduce Cumulative Layout Shift and potentially non-critical CSS depending upon the users theme).

Thoughts? This setting change will allow the user to decide what works best for them rather than providing one single option. This change won't add any significant development time.

Misplon:
Sounds good, thanks.

PR #825, opened 2020-10-26
Add Output Page Builder Layout CSS setting

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/816


Misplon:
Let's consider changing the setting label to:

**Page Builder Layout CSS Output Location**

This is an advanced setting, rather than trying to word this correctly, for the moment, can we consider removing the description?

AlexGStapleton:
That setting name is fine. As for the description, I'm okay with that. It might be worth a dedicated page in the documentation.

Misplon (approved):

gregpriday (approved):

Misplon (approved):

PR #830, opened 2020-12-06
ACF Widgets Fields Compatibility

AlexGStapleton:
This PR adds compatibility for widget fields added by ACF inside of the Page Builder interface. A new action is introduced in this PR called `siteorigin_panels_before_widget_form`. This action is triggered before the widget form is output.

[Test Fields](https://drive.google.com/uc?id=1hFP7XFQcQmX4zt7ZmGT1uGsyKfI26jfN) (to import, navigate to **Custom Fields > Tools**)

### DIFF
--- a/compat/acf-widgets.php
+++ b/compat/acf-widgets.php
@@ -6,6 +6,9 @@ class SiteOrigin_Panels_Compat_ACF_Widgets {
 
 		add_action( 'siteorigin_panels_before_widget_form', array( $this, 'store_acf_widget_fields_instance' ), 10, 3 );
 		add_filter( 'acf/pre_load_value', array( $this, 'load_panels_widget_field_data' ), 10, 3 );
+
+		// Widget Saving inside of Page Builder.
+		add_filter( 'widget_update_callback', array( $this, 'acf_override_instance' ), 10, 4 );
 	}
 
 	public static function single() {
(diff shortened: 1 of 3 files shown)


AlexGStapleton:
There's currently a saving issue present for existing widgets. Once that's resolved I'll submit this for review.

AlexGStapleton:
The saving issue mentioned in my last reply was resolved by https://github.com/siteorigin/siteorigin-panels/pull/830/commits/bf0a26fa906d9d245adfca56f38f3cda8755d243

AlexGStapleton:
You can test this code on the frontend with the following PHP:

```
add_filter( 'siteorigin_panels_the_widget_html', function( $html, $the_widget, $args, $instance ) {
	if ( get_class( $the_widget ) == 'SiteOrigin_Widget_Slider_Widget' && isset( $instance['acf'] ) ) {
		$html = '<ul>';
		foreach ( $instance['acf'] as $key => $value ) {
			$html .= '<li>Field: ' . esc_attr( $key ) . ' . Value: ' . esc_attr( $value ) . '</li>';
		}
		$html .= '</ul>';
	}

	return $html;	
}, 10, 4 );
```

The above PHP will override the output of the Slider widget and output all ACF fields and their values.

gregpriday (approved):

PR #837, opened 2020-12-25
Layout Block: Hide preview button when no preview is available

AlexGStapleton:
This PR will prevent the SiteOrigin Layout Preview button from appearing when in edit mode if there's no panelsData. It'll also reset the panelsData if the user removes all rows from the SiteOrigin Layout to prevent the preview (no longer valid) preview from showing.

This change also prevents a previewing error that commonly occurs while using PHP 8. When using PHP 8, when the SiteOrigin Layout triggers an update to the server and you're in editing mode, clicking the block preview won't do anything - a second click after the preview has loaded is required to view it. Normally it switches to preview mode and adds a loading indicator.

This PR can be triggered by adding a SiteOrigin Layout Block. The preview won't appear. Add a widget and it'll appear. Remove it, and it'll disappear.

### DIFF
--- a/compat/js/siteorigin-panels-layout-block.js
+++ b/compat/js/siteorigin-panels-layout-block.js
@@ -244,14 +244,14 @@ function (_wp$element$Component) {
       };
 
       if (this.state.editing) {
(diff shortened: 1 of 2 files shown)


Misplon:
Nice, dig it.

Misplon (approved):

gregpriday (approved):

PR #843, opened 2021-01-26
Limit CSS Output Location Setting to Classic Editor

AlexGStapleton:
Related https://siteorigin.com/thread/having-issue-with-page-builder-layout-css-output/

The CSS Output Location setting isn't compatible with the Block Editor in a reliable manner - it's rare that it does work. This PR will limit this setting to the Block Editor and add it mentions this limitation in the description in the CSS Output Location description.

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -25,14 +25,23 @@ class SiteOrigin_Panels_Renderer {
 		if ( is_null( $this->inline_css ) ) {
 			// Initialize the inline CSS array and add actions to handle printing.
 			$this->inline_css = array();
-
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
For support related to this limitation, we should recommend [Autoptimize](https://wordpress.org/plugins/autoptimize/).  Autoptimize will indirectly allow this by enabling the **Aggregate CSS Files** **and Also Aggregate Inline CSS** settings. ([setting descriptions](https://kinsta.com/blog/autoptimize-settings/#aggregate-css-files)).

gregpriday (approved):

PR #860, opened 2021-04-24
Change Instant Open Widget default setting

AlexGStapleton:
This will only affect new installs.


Misplon:
@gregpriday Let me know if you have any thoughts on this. I think widgets should open when inserted by default.

Misplon (approved):

gregpriday (approved):

PR #888, opened 2021-06-13
Don't Override Default Row Values

AlexGStapleton:
[Reported here](https://siteorigin.com/thread/default-column-count-editor-add-new-row/)

This PR will prevent PB from overriding the default Row Values.

Test snippet:

```
add_filter( 'siteorigin_panels_row_column_count_input', function( $input ) { 
	return '<input type="number" min="1" max="12" name="cells" class="so-row-field" value="4" />';
} );
```

This will change the default column count to 4.

### DIFF
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -132,7 +132,7 @@ module.exports = panels.view.dialog.extend({
 			this.$('.so-buttons .so-delete').remove();
 		}
 
-		if (!_.isUndefined(this.model)) {
+		if ( ! _.isUndefined( this.model ) && this.dialogType == 'edit' ) {
 			// Set the initial value of the
 			this.$( 'input[name="cells"].so-row-field' ).val( this.model.get( 'cells' ).length );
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. While the field count is correct, the actual row count isn't matching.

Edit_Page_‹_Corp_—_WordPress

AlexGStapleton:
That's actually expected. This filter only adjusts the input, and it doesn't adjust anything else. Its intentions appear as an easy way of quickly adjusting the number of cells used. I can extend it to allow for it to control the default number of columns, but only the default column sizing will be used (100 / number of columns) so it'll be pretty limiting.

I think introducing a new filter that allows you to set the default number of columns and their specific sizes is the way to go. This will retain the original goal of the existing filter, and provide a method of overriding the default columns for those who want it to. We could call the new filter `siteorigin_panels_row_default_columns`. Thoughts?

Misplon:
Sounds good, let's proceed as suggested. Can you update the forum user when able, thanks.

Misplon (approved):

gregpriday (approved):

PR #889, opened 2021-06-13
5.8 New Widgets Interface Compatibility Changes

AlexGStapleton:
This PR:

- Allows the Layout Builder to be edited.

Previously an error message stating that the widget must be edited via an admin interface would be displayed.

- Changes how the Layout Builder `panels_data` is loaded to resolve an incompatibility with new widgets rendering.

This change is in line with how the WB Builder field handles data so PB is set up to handle this sort of usage already.

This change that the Layout Builder needs to be tested (specifically, the act of loading page builder data after a save) in the old Widgets interface, new widgets interface, SiteOrigin Layouts Block and full SiteOrigin Page Builder interface. This change also needs to be tested on pre-WP 5.8.

To test this change, add a Layout Builder, and add any widget to the Layout Builder. Save. Reload the editor and then attempt to access the added widget.

- Post Loop Preview**

The Post Loop needs to be tested in both the new widgets interface preview, and the SiteOrgin Layouts Block Preview.

---

Some notes:
The Template drop down in the post loop widget has different styling but that field is actually being added by WB so it'll be handled in [this PR](https://github.com/siteorigin/so-widgets-bundle/pull/1332).

It's possible the legacy PB widgets aren't compatible with the new widgets interface. We should recommend users change to our more modern widgets or suggest installing the [Classic Widgets](https://wordpress.org/plugins/classic-widgets/) plugin.

### DIFF
--- a/compat/layout-block.php
+++ b/compat/layout-block.php
@@ -32,7 +32,7 @@ class SiteOrigin_Panels_Compat_Layout_Block {
 	}
 	
 	public function enqueue_layout_block_editor_assets() {
-		if (  SiteOrigin_Panels_Admin::is_block_editor() ) {
+		if ( SiteOrigin_Panels_Admin::is_block_editor() || is_customize_preview() ) {
 			$panels_admin = SiteOrigin_Panels_Admin::single();
 			$panels_admin->enqueue_admin_scripts();
 			$panels_admin->enqueue_admin_styles();
--- a/css/admin.less
+++ b/css/admin.less
@@ -2774,8 +2774,10 @@
 		}
 
 		small {
-			color: #333;
-			font-size: 10.5px;
+			color: #999;
+			font-size: 11.7px;
+			font-style: italic;
+			margin-top: .2em;
 		}
 
 		.siteorigin-widget-field {
--- a/css/admin.less
+++ b/css/admin.less
@@ -2766,43 +2766,39 @@
 }
 
 /* 5.8 Widgets Page */
-.widgets-php.block-editor-page {
-	.siteorigin-widget-content {
+.widgets-php.block-editor-page .wp-block-legacy-widget__edit-form .siteorigin-widget-content {
 
-		label {
-			line-height: 1.5;
-		}
-
-		small {
-			color: #999;
-			font-size: 11.7px;
-			font-style: italic;
-			margin-top: .2em;
-		}
-
-		.siteorigin-widget-field {
-			margin-top: 0;
-		}
-
-		select.siteorigin-widget-field {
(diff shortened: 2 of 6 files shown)


AlexGStapleton:
@Misplon I wasn't able to find any other compatibility issues. Let me know if you find any.

Misplon:
WP 5.7.2 using this branch in a build:

* Widget area
* PB Layout Builder
* Layout Block - Layout Builder

If you insert an Archives widget in the above situations, there is no problem. If you insert the Consultant prebuilt layout into the Layout Builder the Layout Builder doesn't retain its contents after save.

AlexGStapleton:
The last commit restored compatibility with 5.7.2 for me. Let me know how you go with it.

Misplon:
* When inserted as a widget and not via the Layout Builder the Post Loop widget Posts heading isn't opening in 5.8 at Appearance > Widgets.
* There is a small alignment issue as follows:

post


We should also adjust the size of the Help link, it looks too big.

Misplon:
@AlexGStapleton I upgraded to 5.8 with a Layout Builder in place, I'm seeing:

122208182-48b8e500-cea3-11eb-8d4a-193f7b2e85dd

AlexGStapleton:
@Misplon 

> I upgraded to 5.8 with a Layout Builder in place, I'm seeing:

I haven't been able to replicate this issue - I tried inserting two Consultant Layouts using a single Layout Builder.

Can you install the Classic Widgets plugin and then send me an export of the layout builder?

> Can you confirm we hit this styling issue in whichever branch is best suited:

No styling adjustments were made prior to the last two.

The pictured fields are added by WB so they would need to be corrected in the WB PR.

The last commit will correct the styling of PB widgets and that'll indirectly help some of the WB fields. Some testing with WB/PB selectively disabled will be needed for the compatibility PR. 

> There is a small alignment issue as follows:

Sections are added by WB. This looks to be present in 5.7.2.



Can you please log this in a separate issue?

> We should also adjust the size of the Help link, it looks too big.

I've adjusted the sizing of the help link to be the same as it was in 5.7.2.

Misplon:
Looking good now, thanks, mate. All ok from my side for now.

Misplon (approved):

AlexGStapleton:
Updated the PB Widgets Page changes to reflect recent Legacy Widgets Block changes.

Misplon:
Super, thanks mate, I've left some feedback in Slack. If you can replicate then perhaps we can make progress on those items.

Misplon (approved):

gregpriday (approved):

Issue #874, opened 2021-05-25
Mobile Editing: Change text links to icons when suitable


Misplon:
Change widget text action links to icons when text links are no longer usable as per the below image:

<img width="182" alt="2021-04-20_03-07-34-1459" src="https://user-images.githubusercontent.com/789159/119530354-360b2e80-bd83-11eb-8831-8992adc32d16.png">

Misplon:
Edit: https://fontawesome.com/icons/edit?style=regular
Duplicate: https://fontawesome.com/icons/copy?style=regular
Delete: https://fontawesome.com/icons/trash-alt?style=regular

Misplon:
@AlexGStapleton The icons from Ric are below. Thanks.

[Icons.zip](https://github.com/siteorigin/siteorigin-panels/files/6654562/Icons.zip)

PR #896, opened 2021-06-18
Cell Edit Icons

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/874

This PR adds cell edit icons that appear when the cells are too small to allow for the text without expanding to a different line.
Please note that this PR will require a build to test.

Before:

After:


Misplon:
Thanks, mate. Can you check the Live Editor? That'll probably be where these are seen the most.

AlexGStapleton:
It does get a bit hard to use when there's a large number of cells so I've added an additional collapse point for both the Live Editor and standard builder. Please note that rows will be larger in the live editor when the smaller collapse point is being used - this is to ensure all icons are visible.





Oh, and I had forgotten to mention but this PR will require a build - I've amended the original PR comment to mention this.

AlexGStapleton:
It's possible for all of the options (text, icons, icons small) to be used in a row:

https://user-images.githubusercontent.com/17275120/122682539-612b5580-d23d-11eb-8890-3408696fa147.mp4

Misplon:
Super, thanks. Awesome improvements, definitely making progress. I uploaded a build to the demo server. Can you perhaps take a look and let me know if we can do anything about the below?

Screenshot 2021-06-21 at 12 05 52

Screenshot 2021-06-21 at 12 06 22

AlexGStapleton:
I've had a look at the demo site and it appears to be using the version prior to the last commit.

Here's the (privately published) page I was testing:
https://demo.siteorigin.com/north/wp-admin/post.php?post=575&action=edit

Here's what the archives widget of the right of the first looks like for me on hover:


Here's the same row/widget on the demo site:


I went to upload a build to the demo site but for whatever reason I can't actually delete plugins. Can you please try [this build](https://drive.google.com/uc?id=18LPiz5feEi5yzPjR4qbGBXTj6exmqix-)? I've tested this build locally and can confirm it works as expected for me.

Misplon:
Thanks, mate. Can we try address the below?

Screenshot_2021-06-21_at_15_11_32

AlexGStapleton:
Are you using the provided build? Here's what I see when I hover that widget:

Misplon:
Yeah, you can see the build is installed if you head over to plugins. Let me try incognito.

AlexGStapleton:
The last commit will apply the styling of the small icons to all icons when in the live editor. This should ensure it's always easier to access the icons.



I'm going to increase the collapse point of the small icons as they don't appear to be appearing as often as they should.

AlexGStapleton:
Icons will now appear when the cell is 215px wide (was182px). The smaller icons will appear when the cell is 125px wide (was 95px). 
@Misplon Feel free to fine tune these amounts, just note that you'll need to either make a build or be running `build:dev` to see a difference.

Misplon (approved):

gregpriday (approved):

Issue #679, opened 2019-06-17
Forminator: Load form using AJAX with Yoast SEO breaks shortcode output


Misplon:
To recreate the conflict:

1. Activate Yoast SEO.
2. Activate Forminator, create a form, edit form settings, in the Behavior section scroll down to Rendering and enable Load form using AJAX.
3. Insert the form into Page Builder using the form shortcode.

In this configuration, the form doesn't load. If Yoast SEO is deactivated or if Load form using AJAX is disabled, the form will load. No console or PHP errors could be found. I couldn't find any specific Yoast SEO settings that changed the outcome.

Misplon:
Checked. This issue remains the same 27th July 2019.

AlexGStapleton:
This issue is caused by Yoast loading a copy of the pages content to generate the Open Graph description. Specifically, this line:

https://github.com/Yoast/wordpress-seo/blob/2e6b272ad62001a05ca113e35a1205376058f9c4/src/presentations/indexable-post-type-presentation.php#L191

That line triggers the_content and Forminator likely has a check that prevents it from running twice and that's causing issues. I'll investigate Forminator's side of things shortly to see if this check can be accounted for.

It should be noted that this issue can be avoided by manually setting a meta description.

AlexGStapleton:
I've looked into this further and I don't see a simple way for us to resolve this issue without doing somethings we ideally shouldn't do (like removing their shortcode briefly and re-adding it, or forcing their assets to output twice). We should recommend users either set a manual meta description or disable Ajax Loading.

Misplon:
Thanks, mate. Do you perhaps want to add a paragraph here for us? https://siteorigin.com/page-builder/troubleshooting-problems/.

AlexGStapleton:
How does that look?

Misplon:
Sorted. Thanks.

AlexGStapleton:
Re-opening due to https://github.com/siteorigin/siteorigin-panels/pull/910

PR #910, opened 2021-07-29
Yoast OpenGraph: Bypass PB As Needed To Improve Compatibility

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/679 & Ninja Form reCAPTCHA error.

When a meta description isn't set, Yoast will render a version of the page to generate the excerpt using `get_the_excerpt`. This is the correct way of doing this, but it can be problematic when plugins account for situations where their plugin is rendered multiple times this which Forminator and Ninja Form both do . This is also compounded by the fact that Page Builder will override `the_content` to replace the fallback content with the generated page builder layout. This PR resolves this entire issue by selectively disabling Page Builder when Yoast is generating the OpenGraph meta description.

To test this PR please set up Forminator form as outlined in #679. Ninja Forms can also be tested but it will require setting up reCAPTCHA so testing it is a bit more involved - I should note that the premium user who reported this conflict has confirmed that fix has worked.

On merge, remove [the Forminator paragraph on this page](https://siteorigin.com/page-builder/troubleshooting-problems/).

### DIFF
--- a/compat/yoast.php
+++ b/compat/yoast.php
@@ -1,4 +1,23 @@
 <?php
+/**
+ * If Yoast OpenGraph is enabled, we'll need disable PB when it gets the excerpt 
+ * to avoid conflicts with other plugins.
+ *
+ */
+function siteorigin_yoast_opengraph_panels_disable( $content ) {
+	global $wp_current_filter;
+	if ( count( $wp_current_filter ) > 2 && $wp_current_filter[1] == 'wpseo_head' ) {
+		// Temporarily disable Page Builder for this instance of the_content.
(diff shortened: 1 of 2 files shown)


AlexGStapleton:
@Misplon I came across this fix when looking into a workaround for Ninja Form reCAPTCHA (instead of the reCAPTCHA you would get "reCAPTCHA is not valid" message). Nina Form provided some great feedback that lead to me thinking of this solution. This PR will also help with other plugins that are affected by this issue that we're simply not aware of.

Misplon (approved):

gregpriday (approved):

Issue #866, opened 2021-05-10
Using CSS to Prevent Container Breaker CLS


AlexGStapleton:
We've received a few reports of the container breaker causing [CLS](https://web.dev/cls/) issues. This is due to the unavoidable delay before the container breaker JavaScript is run simply due to how the DOM works - it's out of our hands.

There is, however, a way to avoid the CLS and that's by using CSS instead of JavaScript. Basically, we remove the theme container globally and then apply it as needed. This removes the need for the container breaker code so there's no "jump" shortly after the page starts loading as a result of it.

### Settings
To accomplish this using CSS, we'll need to introduce two new settings to the Page Builder Layouts tab. These settings will be quite technical as there's no way for us to automate them as automating them will result in a CLS impact. These settings are:

**Theme container selector**
The setting is used for overriding the theme container. This is done on load using CSS rather than CLS and it's going to be done with a high priority `wp_head`.

**Theme Container Width**
This setting is the exact width of the theme container. This cannot be calculated on the fly using the theme container selector as that'll result in a CLS due to the delay before it's able to be calculated and applied.

If both of the above settings are applied, we'll remove the container breaker JS (this would technically resolve https://github.com/siteorigin/siteorigin-panels/issues/557) and manually apply row sizing as needed using CSS.

### How We'll Use Those Settings to Avoid the CLS

For rows that are set to Full Width Stretched, no container is applied.

For full width rows, we apply the container width to either `.panel-row-style`, or a container inside of `.panel-row-style` (using `siteorigin_panels_inside_row_before` hook). This is the most complicated Row Layout and will require the most overall testing.

For standard rows, the container width is applied to `.panel-grid`.

The container width is always applied using `max-width` to avoid potential responsive issues.

### Possible Hooks for Themes

We could allow themes to pre-fill container settings. This will remove the technical concerns and make the process much simpler overall. Those fields could be called: `siteorigin_panels_default_theme_container_selector` and `siteorigin_panels_default_theme_container_width` 

### Final Thoughts

The CSS approach is the best way to resolve the CLS reports. This is, however, quite a technical change and there's no real way around that. There will be older themes that are simply not compatible with it, and there's no real way for us to get around that potential situation outside of recommending users change to a more modern theme.

AlexGStapleton:
@Misplon Let me know if you have any questions or suggestions.

AlexGStapleton:
I've spoken to @Misplon about this over Slack.

[...]

PR #911, opened 2021-08-01
(Optional) Container Breaker Using CSS Rather Than JS

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/866

This PR introduces the two filters outlined in the initial PR. It no longer includes settings as it has been decided this change is a little bit too tricky to manage without assistance. To test this PR please configure the 

`siteorigin_panels_theme_container_width` filter (int) and `siteorigin_panels_theme_container_selector` (CSS selector). For example, here's a snippet for Corp:

```
add_filter( 'siteorigin_panels_theme_container_width', function( $container ) {
	return '1170px';
} );

add_filter( 'siteorigin_panels_theme_container_selector', function( $selector ) {
	return '.site-content .corp-container';
} );
```

To test this please add four rows with each of the row layouts selected and confirm:

- Standard row is the expected width.
- Full Width has a special container wrapper applied inside of the row.
- Full Width rows still have a full width background.
- Full Width Stretch doesn't have any container applied at all.

([here's a test layout](https://drive.google.com/uc?id=1t8KdPQXSeZ4LcaZqJ3v-p5YmM23oyg-r))

### DIFF
--- a/inc/renderer.php
+++ b/inc/renderer.php
@@ -3,6 +3,7 @@
 class SiteOrigin_Panels_Renderer {
 
 	private $inline_css;
+	private $container;
 
 	function __construct() {
 		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_styles' ), 1 );
@@ -75,6 +76,10 @@ class SiteOrigin_Panels_Renderer {
 		if ( empty( $layout_data ) ) {
 			$layout_data = $this->get_panels_layout_data( $panels_data );
 			$layout_data = apply_filters( 'siteorigin_panels_layout_data', $layout_data, $post_id );
+		}
+
+		if ( empty( $this->container ) ) {
+			$this->container = SiteOrigin_Panels::container_settings();
 		}
 
 		// Get some of the default settings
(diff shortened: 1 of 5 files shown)


AlexGStapleton:
I think we should remove the Page Builder settings and rely on filters instead. A user that is using this build has reported a few alignment issues (this was expected). I've come across a situation where a few things need to be accounted for but there's really no way we could account for those situations without adding a ton of checks which I don't think we should do as it sort of goes against the "simpler" CSS approach. Instead, I think they should be accounted for by CSS snippets.

North with Sharedaddy, and Jetpack Related Posts installed:

```
.entry-header,
.entry-content footer.entry-footer,
.sharedaddy,
#jp-relatedposts.jp-relatedposts,
.site-main .post-navigation {
	max-width:1040px;
	margin:0 auto;
}
```

Misplon (approved):

AlexGStapleton:
The last commit will now add a body class when the CSS Container Breaker is active. This will allow for themes (and us) to override things only when the container breaker is active. These means the previously referenced CSS is now:

```
body.siteorigin-panels-css-container .entry-header,
body.siteorigin-panels-css-container .entry-content footer.entry-footer,
body.siteorigin-panels-css-container .sharedaddy,
body.siteorigin-panels-css-container #jp-relatedposts.jp-relatedposts,
body.siteorigin-panels-css-container .site-main .post-navigation {
	max-width:1040px;
	margin:0 auto;
}
```

Misplon (approved):

AlexGStapleton:
https://github.com/siteorigin/siteorigin-panels/pull/911/commits/43525c704ed66e28c3ce8b23ae260c39dec7e7d6 resolves a column sizing issue on mobile (it's visible in the last screenshot of [this comment](https://github.com/siteorigin/siteorigin-panels/issues/866#issuecomment-917472002)). To test it:

- Add a row with three columns and set the row to Full Width.
- Add an Archives widget to each of the columns and give them a background colour.
- View the row on mobile and confirm the sizing issue.
- Pull the latest commits of this PR and view the row again.

AlexGStapleton:
https://github.com/siteorigin/siteorigin-panels/pull/911/commits/22f42fd713cab56a45759ad3ba5896bcf376365b will resolve a display issue with the CSS Container Breaker and stretched columns and a cell with a background applied but no widgets.

It'll also ensure only the needed Cell Vertical Alignment selector is output. Previously, both the styled and unstyled selector was output when only one was ever output.

[Here's a test layout](https://drive.google.com/uc?id=1V2TN2cb7DVJq1BaqdJSbxoeNmD6bhf5h). Please apply a cell background to the last column on the first two rows to test the last commit.

Misplon (approved):

gregpriday (approved):

PR #919, opened 2021-09-06
Links Hover Color

AlexGStapleton:
This PR adds the Links Hover Color setting. Previously, when a user altered the link color, the color wouldn't change on hover and this setting will avoid that situation by allowing them to also set the Links Hover Color.

Links Hover Color is not tied to any setting and can be set independently to the Links Color setting. There are widgets that will have more specific CSS than this setting, and there's no way of really getting around that without potential display issues.

### DIFF
--- a/inc/styles.php
+++ b/inc/styles.php
@@ -347,7 +347,7 @@ class SiteOrigin_Panels_Styles {
 		);
 
 		$fields['link_color'] = array(
-			'name'        => __( 'Links Color', 'siteorigin-panels' ),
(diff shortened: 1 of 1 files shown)


AlexGStapleton:
@Misplon What are your thoughts on Link_s_ in this instance? I called the setting Links Hover Cover due to the base setting being called Links Color I think removing the S in both settings would make sense.

Misplon (approved):

gregpriday (approved):

PR #923, opened 2021-09-15
Make $css public to allow more extensive filtering

bbceg:


AlexGStapleton (approved):
Thanks Conrad. :)

AlexGStapleton:
There's really no reason for $css not to be exposed and changing the visibility of $css won't result in any issues with existing usage.

Issue #962, opened 2022-01-08
Sidebars-Emulator Causes PHP Warning


MachineITSvcs:
I was receiving the following PHP error/warning:

`PHP message: PHP Warning:  Undefined array key 1 in /web/wp-content/plugins/siteorigin-panels/inc/sidebars-emulator.php on line 53`

I believe this is due to the preg_match not finding a match, and thus $num_match is not set a valid value. Considering that preg_match returns 0 on no match or false on failure, may I suggest that the lines 52 and 53 be updated to something like:

```
if(!empty(preg_match('/-([0-9]+$)/', $sidebar_id, $num_match))) {
    $args[0][ $num_match[1] ] = $widget_instance;
}
```

This will require the match to be found before attempting to set to what would otherwise be an undefined variable array value, similar to how you already check $opt_matches is not empty on line 33. Considering I also don't see the $args being set in the function, but passed as an argument, should there be logic beforehand that ensures $args[0] is set as well? Maybe something like:

`if(!isset($args[0])) $args[0] = array();`

Perhaps on line 30 (before the check for $opt_matches), and maybe move the "return $args[0];" line to the end of the function instead of before the last curly bracket "}" unless you don't want this function to return $args[0] when no $opt_matches found?

AlexGStapleton:
Hi @MachineITSvcs,

Thank you for the report. I'm not too sure how `preg_match` could result in no results, but I absolutely can see the possibility of a notice if it did. There's really nothing we can do to emulate the widget if it does return nothing so not setting the id is the way to go. I've accounted for this possibility here: https://github.com/siteorigin/siteorigin-panels/pull/964

I've prepared a build of that PR for you to try, and you can [download it by clicking here](https://drive.google.com/uc?id=104Y5mU7gxrVpfOnssL8sQlqPiA4-O2Cb).

Before proceeding, do a backup.

Navigate to **Plugins > Installed Plugins** and deactivate Page Builder by SiteOrigin, and then delete it. Scroll to the top of the page and click Add New, Upload Plugin and upload siteorigin-panels.2.16.1-beta.zip. When prompted, activate Page Builder by SiteOrigin. Does this notice still occur for you?

> Considering I also don't see the $args being set in the function, but passed as an argument, should there be logic beforehand that ensures $args[0] is set as well?

[...]

PR #964, opened 2022-01-09
Sidebar Emulator: Account for potential notice when setting widget id

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/962

To test this PR please enable the Sidebar Emulator setting, add a bunch of WordPress and SiteOrigin Widgets to the sidebar and if no errors are logged, it worked as expected.

### DIFF
--- a/inc/sidebars-emulator.php
+++ b/inc/sidebars-emulator.php
@@ -50,7 +50,7 @@ class SiteOrigin_Panels_Sidebars_Emulator {
 						if ( $instance_class == $widget_class && ! empty( $sidebar_id ) ) {
 							//The option value uses only the widget id number as keys
 							preg_match( '/-([0-9]+$)/', $sidebar_id, $num_match );
-							if ( ! empty( $num_match ) && count( $num_match ) > 1  ) {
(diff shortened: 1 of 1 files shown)


Misplon (approved):

MachineITSvcs:
I see your logic here, but you could also just use a !empty($num_match[1]) which would check that the key exists and would return false if $num_match was empty or key 1 was undefined, vs checking $num_match isn't empty and has more than 1 key. I mean regardless we're talking an inconsequential difference in execution time, and perhaps the current conditions improve readability.

MachineITSvcs:
Sorry. I believe the key was again left out. I'd recommend:

`if(!empty($num_match[1])) {`

What do you think?

AlexGStapleton:
Hi,

In this instance, it's preferable that the direct key isn't checked. If there are any matches, there will be two as the first result looks like `-12345` and the second item will always be `12345` (the same value as the first one). Checking the exact array index of 1 will result in l in issues if there aren't any matches unless the check is more specific - `! empty( $num_matches ) && ! empty( $num_matches[0] )`

Kind regards.
Alex

Issue #979, opened 2022-04-02
Layout Block: Export JSON file name


Misplon:
@AlexGStapleton, I remember JSON file names from the Layout Block used the page name. At the moment, the JSON file is named using the page ID. Can you perhaps take a look to see if something changed?

PR #980, opened 2022-04-04
SO Layouts Block: Export Filename Fix

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/979

In a recent WordPress update, the title field was changed from an input to a div so the previous method of getting the post title no longer works. This PR accounts for the new field while also retaining the older Block Editor field and Classic Editor field

### DIFF
--- a/js/siteorigin-panels/dialog/prebuilt.js
+++ b/js/siteorigin-panels/dialog/prebuilt.js
@@ -180,8 +180,8 @@ module.exports = panels.view.dialog.extend( {
 			var $$ = $( this );
 			var panelsData = thisView.builder.model.getPanelsData();
 			var postName = $( 'input[name="post_title"], .editor-post-title__input' ).val();
-			if ( ! postName ) {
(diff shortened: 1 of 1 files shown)


Misplon:
Thanks, mate. I've uploaded a build to https://demo.siteorigin.com/corp/wp-admin/post.php?post=3554&action=edit. The issue doesn't seem to be resolved.

AlexGStapleton:
The linked page is powered by the Classic Editor so it won't be affected by this issue. Regardless, the export of that page is working as expected - I'm getting sandbox-two.json. I've identified a situation where the post title still won't be detected and that's accounted for in the last commit. I've uploaded the below build to the demo server.

[siteorigin-panels.2.16.9-beta-3.zip](https://github.com/siteorigin/siteorigin-panels/files/8426935/siteorigin-panels.2.16.9-beta-3.zip)

Misplon (approved):

PR #982, opened 2022-04-12
Restore New WP Widget Area SO Layouts Block Compatibility & Other Improvements

AlexGStapleton:
In a recent WP update:
- WP stopped outputting wp-data automatically.
- Removed the `core/editor` in certain contexts (mainly the new Widget Area).

This PR accounts for both of these and it adds Layout Editor and History support to the SiteOrigin Layouts Block in the new Widget area - it was a relatively simple change.

This PR modifies core JS files so a build will be required.

### DIFF
--- a/compat/js/siteorigin-panels-layout-block.js
+++ b/compat/js/siteorigin-panels-layout-block.js
@@ -316,13 +316,13 @@ wp.blocks.registerBlockType('siteorigin-panels/layout-block', {
         var isNewWPBlockEditor = jQuery('.widgets-php').length;
 
         if (!isNewWPBlockEditor) {
-          wp.data.dispatch('core/block-editor').lockPostSaving();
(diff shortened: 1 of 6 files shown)


Misplon:
@AlexGStapleton Thanks, mate. Can you check that the SiteOrigin Layout Block in the Block Editor works normally for you? Ideally, an existing SiteOrigin Layout Block with a layout before switching to this branch.

Edit_Page_“Blocked_Up”_‹_Puro_—_WordPress

AlexGStapleton:
I've reverted [Layouts Block: Use Correct Editor API on New Widgets Page](https://github.com/siteorigin/siteorigin-panels/pull/982/commits/cc19a35a9365bf382bbafb2314462f199e2565b4). That looked to be more reliable overall, but turns out it's even more inconsistent - `block-editor` isn't always available in the Block Editor. :/

PR #988, opened 2022-06-29
Styles: Add Toggle Field, and Box shadow & Box shadow hover Settings

AlexGStapleton:
This PR adds a Toggle field for Styles and adds a `Box shadow` and `Box shadow hover` settings. These settings are able to be applied to Rows, Cells, and Widgets. If either Box Shadow is enabled and none of the box shadow settings are adjusted, the `box-shadow` that'll be used is: `0 3px 10px rgb(0 0 0 / 0.2)`.

This PR will require a build to test as core PB JS and CSS is adjusted.


AlexGStapleton:
I've fixed the label and resolved the conflict.

AlexGStapleton:
> Both Disabled and Enabled display for me at the same time:

That's how it was in the reference image - it's to indicate which side is enabled and disabled. Although, in retrospect, the toggled state should be more obvious than the reference image.

Wait... you're not seeing the toggle field. Does [this build ](https://drive.google.com/uc?id=1xRaIDWS4vYApQhZxC462OmbXB6OtG_ag) display a toggle field? (rather than a checkbox) It should look like this:



Please note that the above linked build includes the most recent commits.


**The box-shadow hover effect seems to be present without me enabling that settings section.**
I can't seem to replicate this. If you're still able to replicate this with the latest version of this branch or the provided build, can you please send me a copy of the layout you're using?

AlexGStapleton:
> Which reference image are we using for this?

https://siteorigin.slack.com/archives/D0MGRNNTA/p1647871476550329

Feel free to make any adjustments you would like. If you decide you actually would like how the Toggle Visibility handles things, let me know and I'll make that now.

Oh! That's what it's meant to be. My mistake.

> The actual toggle styling doesn't load for me. Perhaps I need to try a build?

 I'm not too sure why it's not working but yes, please try a build. [Here's a build](https://drive.google.com/uc?id=1D720IgvP-WQdB_27eihK2xaFXoDqsuVu) I tested just now and it's working as expected.

Misplon:
@AlexGStapleton A quick follow-up on this PR when able. Thanks.

AlexGStapleton:
> When enabling hover but making no adjustments, are we loading a default as we do for the main setting? (0 3px 10px rgb(0 0 0 / 20%);)

Yes. Having a different default hover wouldn't be a bad idea. Do you have one in mind?

Misplon:
@AlexGStapleton Moving my Slack feedback here.

- Can we use `left: 9px` for ON.
- Is it possible to use `box-shadow: rgba(0, 0, 0, 0.15) 0px 5px 15px 0px;` as our default. This is the Shopify example from https://getcssscan.com/css-box-shadow-examples.
- For the hover default, how about `box-shadow: rgb(0 0 0 / 30%) 0px 5px 15px 0px;` with a `.3s` transition.

AlexGStapleton:
Ready for review.

Misplon (approved):

PR #992, opened 2022-07-12
Add Jetpack Copy Post Compatibility

AlexGStapleton:
This PR adds [Jetpack Copy Post](https://jetpack.com/support/copy-post/) compatibility by ensuring that Page Builder data is copied when Jetpack copies the page - which doesn't always happen.


AlexGStapleton:
User reported success.

PR #1002, opened 2022-08-18
Settings descriptions updates

Misplon:
@AlexGStapleton This PR makes a small update to the Attributes > Class setting description and removes several Design setting descriptions that are perhaps redundant. The extra space will hopefully make the section easier to navigate once we add the shadow and radius settings. I've moved this chat from Slack. Let me know if you're keen on these changes. No worries either way.


AlexGStapleton (approved):
Missed this over Slack. This is a good issue and I agree with your changes.

Misplon:
Thanks for checking.

Issue #938, opened 2021-10-12
Provide Method for Widgets to Validate Fields


AlexGStapleton:
Currently, all validation is handled when saving the page and that can make it tricky to reject changes and suggest alternatives. We should implement a JavaScript trigger for developers to validate any data (although not change it) before it's saved and allow them to "reject" changes by preventing the widget from being saved until the validation passes.

PR #1007, opened 2022-10-04
Panels `close_dialog_validation`: Client Side Event to Allow for Widget Validation

AlexGStapleton:
This PR resolves https://github.com/siteorigin/siteorigin-panels/issues/938. It's also the Page Builder's side of widget validation for the Widgets Bundle (which will interface with this and a similar event on its side). Please note that this PR modifies core PB JS files so a build is required.

Here's some test JavaScript you can try by adding them to the browser console.

This snippet is a simple way of testing validation success and failure. It causes a 50/50 chance of widget failing validation.
```
jQuery( document ).on('close_dialog_validation', function( e, values, widget, id, instance ) {
	return Math.random() >= 0.5;
} );
```

This snippet applies to the Archives widget and makes the title field required. It also shows an example of indicating an issue by adding a red border around the title field.
```
jQuery( document ).on('close_dialog_validation', function( e, values, widget, id, instance ) {
	var valid = true;

	if ( widget == 'WP_Widget_Archives' ) {
		if ( ! values.title.length ) {
			jQuery( '[name="widgets[' + id + '][title]"]' ).css( 'border', '2px solid #f00')
			valid = false;
		} else {
			jQuery( '[name="widgets[' + id + '][title]"]' ).css( 'border', '1px solid #8c8f94')
		}
	}
	return valid;
} );
```

### DIFF
--- a/js/siteorigin-panels/view/dialog.js
+++ b/js/siteorigin-panels/view/dialog.js
@@ -379,6 +379,29 @@ module.exports = Backbone.View.extend( {
 			silent: false
 		}, options );
 
+		// Allow for external field validation when attempting to close a widget.
+		if ( typeof this.widgetView == 'object' ) {
+			if ( typeof values.widgets == 'object' ) {
+				validSave = $( document ).triggerHandler(
+					'close_dialog_validation',
+					[
+						// Widget values.
+						values.widgets[ this.model.cid ],
+						// Widget Class
+						this.model.attributes.class,
+						// Model instance - used for finding field markup.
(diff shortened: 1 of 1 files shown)


Misplon:
@AlexGStapleton Follow-up on my Slack screencast (uses a build). Perhaps you could send a build from your side? I tried again today but am not winning testing with the above snippets (using a build).

Misplon:
My mistake - didn't hit enter. Looks great.

Misplon (approved):

PR #1017, opened 2022-11-09
Row Color Adjustments

AlexGStapleton:
This PR will allow row colors to wrap and centers them.
Screenshot 2022-11-09 at 16-06-33 Edit Page ‹ SO — WordPress

It also sets the default active color for rows and hover color for cell dividers.

This PR will require a build.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -251,8 +251,8 @@
 						clear: both;
 						display: flex;
 						flex-wrap: wrap;
-						gap: 2px;
-						justify-content: center;
+						gap: 4px;
+						justify-content: space-evenly;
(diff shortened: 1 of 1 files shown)


Misplon:
@AlexGStapleton Comparing the way the colors are displayed with the addon enabled and disabled, I think we should use:

```
 gap: 4px;
 justify-content: space-evenly;
```

Please, give those values a shot and let me know if you agree.

Thanks

AlexGStapleton:
Looks good. 👍

Misplon:
Thanks, mate.

Misplon (approved):

Issue #1014, opened 2022-10-27
Background Image: Introduce Size Options


Misplon:
Conditionally display size options.

PR #1018, opened 2022-11-09
Add Background Image Size Setting & Conditionally Hide Background Related Settings

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/1014

This PR introduces the Background Image Size setting and conditionally hides Background Image related settings (display and size) until a background is set.

This PR will require a build.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -2079,7 +2079,8 @@
 		.style-input-wrapper {
 			.clearfix();
 
-			input {
+			input,
+			select {
 				max-width: 100%;
 			}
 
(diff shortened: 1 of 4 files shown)


Misplon:
@AlexGStapleton Thanks, mate. Good work.

- Can we do anything about the select field overflow issue on smaller screens?
- Is `cover` working for you in the last row here https://demo.siteorigin.com/corp/sandbox/?

Misplon:
A build will need to be re-uploaded to the demo server to check the above-linked page.

PR #1108, opened 2023-10-31
Rename Full Width Stretched Padded Setting

AlexGStapleton:
It wasn't very clear what this setting does so this PR will rename it to "Full Width Stretched With Padding Support". This should make it clear that it supports padding while the base Full Width Stretched layout doesn't.


andrejarh:
This is more understandable, yes. But does the old one has any advantage over the "Full Width Stretched With Padding Support"? 

If not, then maybe the "Full width stretched" could be renamed to "Full width stretched - depreciated" (Or perhaps not even shown at all in the editor, just supported for the sake of backward compatibility)? The one with padding support could be just "Full width stretched"? 

Just an idea...

PR #1297, opened 2025-12-26
Add contextual menu action hooks

Misplon:
## Summary
- add a `siteorigin_panels_contextual_menu_actions` filter to supply custom contextual actions
- support per-action priority ordering so custom actions can be placed before destructive items
- expose a shared `siteoriginPanels.events` bus for contextual menu action handlers
- merge custom actions into row/widget contextual menus and emit `contextual_menu_action` for non-core actions

## Why
- Page Builder doesn’t currently expose an extension point for contextual menu actions
- SiteOrigin Premium needs to add “Save as Custom Layout” without modifying core menus

## Testing
- not run (local build tooling not available)

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -387,36 +387,7 @@ class SiteOrigin_Panels_Admin {
 				return $tab;
 			}, $tabs );
 
-			$contextual_actions = apply_filters(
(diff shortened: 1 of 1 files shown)


Misplon:
Added in support of https://github.com/siteorigin/siteorigin-premium/pull/740.

PR #1301, opened 2026-02-18
Fix Row Preview Percentage Alignment

### LINKED ISSUE #1212: Rewrite Column Preview
We've had trouble with the column preview in the past, and to ensure that doesn't happen again, we'll need to rewrite it.

Misplon:
## Summary
- Rework the row preview weight markup so the numeric value and percent symbol render as one component.
- Replace fragile text/HTML updates with targeted updates to a dedicated `.preview-cell-weight-value` node.
- Center the preview label container with a single positioning model to avoid drift and overflow in different rendering contexts.

## Issue
This PR fixes: #1212
Issue URL: https://github.com/siteorigin/siteorigin-panels/issues/1212

## Testing
- Open Row Edit dialog and verify preview labels remain centered.
- Verify `%` remains inside the preview box during drag and manual input updates.
- Verify clicking `%` still focuses the editable value/input.

### DIFF
--- a/css/admin.less
+++ b/css/admin.less
@@ -1713,7 +1713,7 @@
 				}
 
 				.row-cell-column input {
-					width: 58px;
+					width: 100%;
 				}
 
 				button {
@@ -1726,6 +1726,69 @@
 
 				.row-cell-column {
 					flex: 1;
+					display: flex;
+					align-items: center;
+					gap: 8px;
+
+					.so-row-count-control {
+						display: inline-block;
+						overflow: hidden;
+						position: relative;
+						width: 58px;
+						border-radius: 0;
+					}
+
+					.so-row-count-stepper {
+						align-items: center;
+						background: #f6f6f6;
+						border-radius: 0;
+						display: flex;
+						flex-direction: column;
+						height: calc( 100% - 2px );
+						justify-content: center;
+						position: absolute;
+						right: 1px;
+						top: 1px;
+						width: 18px;
+						z-index: 3;
+						overflow: hidden;
+					}
+
+					.so-row-count-step {
+						align-items: center;
+						background: transparent;
+						border: 0;
+						color: #5e6d72;
+						cursor: pointer;
+						display: inline-flex;
+						font-size: 12px;
+						font-weight: bold;
+						height: 50%;
+						justify-content: center;
+						line-height: 1;
+						margin: 0;
+						padding: 0;
+						width: 100%;
+						margin-top: 0;
+						border-radius: 0;
+
+						&:hover {
+							color: #1d2327;
+						}
+					}
+
+					.so-row-field {
+						-moz-appearance: textfield;
+						padding-right: 18px;
+						border-radius: 0;
+						margin: 0;
+
+						&::-webkit-outer-spin-button,
+						&::-webkit-inner-spin-button {
+							-webkit-appearance: none;
+							margin: 0;
+						}
+					}
 				}
 
 				.cell-resize-container {
--- a/css/admin.less
+++ b/css/admin.less
@@ -1736,6 +1736,10 @@
 						position: relative;
 						width: 58px;
 						border-radius: 0;
+
+						&:focus-within .so-row-field {
+							border-color: #3c434a;
+						}
 					}
 
 					.so-row-count-stepper {
@@ -1994,6 +1998,10 @@
 							overflow: hidden;
 							position: relative;
 							width: 80px;
+
+							&:focus-within .preview-cell-weight-input {
+								border-color: #8c8f94;
+							}
 						}
 
 						.preview-cell-stepper {
--- a/css/admin.less
+++ b/css/admin.less
@@ -1907,17 +1907,18 @@
 						}
 
 						.preview-cell-unit {
-							left: 10px;
-							position: absolute;
-							top: -1.5px;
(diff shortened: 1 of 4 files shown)


AlexGStapleton:
I'm getting the below result. Is this happening for you? If not, can you please send me a build you've confirmed?

Open the row (no clicks):
image

I'm also getting [slight movement when clicking in and out of the column width field](https://drive.google.com/uc?id=1BRISkzC36Or4gIIaqZW38iLTq2_5Uq4o).

Misplon:
Here is a build to try:

[siteorigin-panels.fix-preview-number-centering.zip](https://github.com/user-attachments/files/25576959/siteorigin-panels.fix-preview-number-centering.zip)

AlexGStapleton:
The first issue I reported is not occurring with that build. Unsure what happened with the build process on my end.

The second issue is worse when using the provided build. Here's a video of it:

https://github.com/user-attachments/assets/4f23720f-c5ab-42d4-b688-da08f926ead0

Misplon:
Thanks, I was ok with that but have gone a new route now with a follow-up PR to get Column Count to match the custom controls.

Misplon:
Follow-up https://github.com/siteorigin/siteorigin-panels/pull/1301.

AlexGStapleton:
Provided feedback to the custom stepper in https://github.com/siteorigin/siteorigin-panels/pull/1314
Once those issues are resolved in both PR's, I think both are good to merge.

AlexGStapleton (changes requested):
That last comment was meant to be the review - sorry about that.

Misplon:
Quick follow-up for final confirmation:

- The outstanding feedback you left on the custom stepper in #1314 has now been fully addressed (including the final stepUp/stepDown simplification point).
- For #1301, we re-checked the row preview width-control changes and don’t see any additional unresolved code-level asks beyond the focus/movement behavior thread already discussed.

Could you confirm whether you still see the input movement issue on the latest `codex/fix-preview-number-centering` branch, or if this is now good to merge from your side?

AlexGStapleton:
I can confirm that part works as expected and I have no feedback. The outstanding issue related to the steppers which is directly linked to https://github.com/siteorigin/siteorigin-panels/pull/1314.

Both this PR and that PR are using the same stepper code (just in two locations). That PR has had its markup updated, but not this one (#1301). This can't be merged until the underlying updated code has been copied over, and ideally, that shouldn't happen until https://github.com/siteorigin/siteorigin-panels/pull/1314 is merged (saves having to update two sets of changes).

Misplon:
This PR now inherits the finalized column count stepper implementation from #1314, which was the remaining blocker Alex identified in his March 8 review.

**What was ported from #1314:**
- `role="group"` with `aria-label` on the row-set-form wrapper
- Proper `<label>` element with `for` attribute for "Column Count"
- Screen-reader instruction text and `aria-live="polite"` on the input
- Integrated stepper buttons using native `stepUp()`/`stepDown()`
- Corresponding CSS for the stepper control layout

**Additional fix in this commit:**
- Removed `tabindex="-1"` from the preview cell stepper buttons, matching the same accessibility pattern Alex flagged in #1314

The original preview alignment fix remains intact — percentage values stay centered, `%` stays inside the preview box, and clicking/focusing the field does not introduce extra movement.

@AlexGStapleton ready for review once #1314 is merged.

Misplon:
@AlexGStapleton Quick clarification on how the branches ended up after your March 8 feedback.

`#1301` now contains both:
- the original row preview alignment/focus fix, and
- the column-count control updates from `#1314` (ported in `6330c231`), plus the preview stepper accessibility/focus-blur follow-up in `16585b37`.

So the main PR now includes the modernized Column Count control as well as the preview-control changes.

Given that, my suggestion would be:
- merge `#1301` if the combined result looks correct from your side, and
- close `#1314` as redundant, since its column-count changes are now already present in `#1301`.

If you would prefer these to stay separate instead, we can unwind that and handle them independently. I wanted to call out the current state explicitly so we don't create confusion or merge the same control changes twice.

AlexGStapleton:
> If you would prefer these to stay separate instead, we can unwind that and handle them independently

No, keep it as is. Reverting changes will just complicate things further.

AlexGStapleton (changes requested):
Remaining issues:
- I'm now having trouble tabbing to the column count stepper.
- The column count stepper doesn't have focus styling anymore.
- It appears the focus styling of the column width stepper has regressed. The styling should be 1:1 with the hover styling.

Misplon:
Thanks for the feedback. Keen to wrap this one up.

We don't use a blue focus border in this modal other than for the col count field, it seems out of place/doesn't match. I've made a series of small refinements.

Misplon:
@AlexGStapleton Can you take a look for us when able? I'll put out a release once this PR is merged. Thanks.

AlexGStapleton:
> We don't use a blue focus border in this modal other than for the col count field, it seems out of place/doesn't match. I've made a series of small refinements.

That's fair. It doesn't directly have to be a border (although that's ideal), it can be a background change.

AlexGStapleton (approved):
Thanks. It's now working as expected, and everything else looks good. 

There's still room for improvement with the focus/hover styling on the column count toolbar, but that last change resolved the only remaining blocker. Good to merge.

PR #1304, opened 2026-02-18
Fix Events Manager Publish Status Regression

Misplon:
## Summary
- Fix a publish regression for Events Manager event post types when Page Builder "Copy Content" is enabled.
- Avoid running a secondary `wp_update_post()` save cycle for Events Manager events during Page Builder content sync.
- For `event` and `event-recurring` post types, sync `post_content` directly to the posts table and clear post cache.

## Problem
When duplicating/editing an Events Manager event with Page Builder data present, publishing from the single event edit screen could revert to Draft during Page Builder's copied-content sync.

## Issue
User-reported bug (no matching open GitHub issue found at the time of PR creation).

## Testing
- Duplicate an Events Manager event containing Page Builder data.
- Change date/time and publish from the single event edit screen.
- Confirm status remains `publish`.
- Confirm behavior is unchanged for non-Events-Manager post types.

### DIFF
--- a/compat/events-manager.php
+++ b/compat/events-manager.php
@@ -24,6 +24,26 @@ class SiteOrigin_Panels_Compat_Events_Manager {
 		add_action( 'em_event_duplicate_pre', array( $this, 'duplicate_pre' ) );
 		add_filter( 'em_event_get_event_meta', array( $this, 'filter_duplicate_meta' ) );
 		add_filter( 'em_event_duplicate', array( $this, 'duplicate_copy_panels_data' ), 10, 2 );
+		add_filter( 'siteorigin_panels_copy_content_update_method', array( $this, 'copy_content_update_method' ), 10, 4 );
+	}
+
+	/**
+	 * Events Manager can force event status to draft during wp_update_post content sync.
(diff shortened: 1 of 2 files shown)


AlexGStapleton (changes requested):
This code should largely be stored in `compat/events-manager.php`.

With that said, if the goal is to prevent external plugins from triggering with this sort of change, it might be worth just using the compatibility code in general. Using wpdb directly will skip any chance of this ever conflicting with any other plugins in the future. Might be a bit much though.

AlexGStapleton (approved):
Looks good. It might be worth moving the `direct_db` code to a dedicated class method (it's quite bulky). Outside of that, good to merge.

Misplon:
@AlexGStapleton I implemented your suggestion from the approved review.

I moved the `direct_db` content sync logic into a dedicated method (`copy_content_update_post_direct_db`) in `SiteOrigin_Panels_Admin` and replaced the inline block with a method call. This keeps `save_post()` easier to read while preserving the same behavior for Events Manager compatibility.

Thank you for the review.

AlexGStapleton (approved):

PR #1309, opened 2026-02-21
Generalize admin asset conflict handling for compat modules

Misplon:
## Summary
- add SiteOrigin_Panels_Compat_Admin_Asset_Conflicts as a central admin_enqueue_scripts compatibility manager
- expose generic filters for conflict handles and excluded screen matchers
- migrate Ditty compatibility to register handles/screen exclusions through the generic filters instead of direct dequeue logic

## Testing
- php -l compat/admin-asset-conflicts.php
- php -l compat/ditty.php
- php -l inc/compatibility.php

### DIFF
--- a/compat/admin-asset-conflicts.php
+++ b/compat/admin-asset-conflicts.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * Generic admin asset conflict handling for Page Builder screens.
+ */
+class SiteOrigin_Panels_Compat_Admin_Asset_Conflicts {
+	public static function single() {
+		static $single;
+
+		return empty( $single ) ? $single = new self() : $single;
+	}
+
+	public function __construct() {
+		add_action( 'admin_enqueue_scripts', array( $this, 'dequeue_conflicting_assets' ), 1000 );
+	}
+
+	/**
+	 * Dequeue conflicting plugin assets on Page Builder screens.
+	 *
+	 * @param string $hook_suffix Admin hook suffix.
+	 *
+	 * @return void
+	 */
+	public function dequeue_conflicting_assets( $hook_suffix ) {
+		if ( ! $this->is_siteorigin_builder_screen( $hook_suffix ) ) {
+			return;
+		}
+
+		$screen = get_current_screen();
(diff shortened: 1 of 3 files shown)


compat/admin-asset-conflicts.php
```diff
@@ -0,0 +1,127 @@
...
+			}
+			wp_dequeue_style( $handle );
+		}
+
+		foreach ( $this->sanitize_handles( $script_handles ) as $handle ) {
+			if ( wp_script_is( $handle, 'enqueued' ) ) {
+				$removed['scripts'][] = $handle;
+			}
+			wp_dequeue_script( $handle );
+		}
```
AlexGStapleton, line 72: I think the dequeue should only occur if the script is found. With this in mind, it might be worth also checking if the script is one of these:
- registered
- enqueued
- queue

compat/admin-asset-conflicts.php
```diff
@@ -0,0 +1,127 @@
...
+
+		$GLOBALS['SITEORIGIN_PANELS_ADMIN_COMPAT']['removed'] = $removed;
```
AlexGStapleton: This should be a public SiteOrigin_Panels_Compat_Admin_Asset_Conflicts variable.

AlexGStapleton (changes requested):

Misplon:
@AlexGStapleton I applied the requested changes.

I updated the dequeue flow to only dequeue when a handle is available in a valid WordPress state (`registered`, `enqueued`, or `queue`) for both styles and scripts.

I also changed the compatibility debug state to use a public class-named global variable: `$GLOBALS['SiteOrigin_Panels_Compat_Admin_Asset_Conflicts']`.

Thank you for the review.

AlexGStapleton:
> I also changed the compatibility debug state to use a public class-named global variable $GLOBALS['SiteOrigin_Panels_Compat_Admin_Asset_Conflicts'].

Globals should be actively avoided. Please use a publicly exposed class property. [Effectively, what was done here, just public](https://github.com/siteorigin/siteorigin-panels/pull/1307/changes/a0cebb24f92114ec0fceda7c58788409f44b54f7).

Outside of that, the last change looks good. Once a public class property is used rather than a global is used, please feel free to merge.

PR #1310, opened 2026-02-23
Fix WooCommerce Shop Page Context Handling

Misplon:
## Summary
- Scope SiteOrigin Page Builder WooCommerce content rendering to the Shop archive description context only.
- Avoid intercepting unrelated `woocommerce_format_content` calls (for example, checkout Terms and Conditions content).
- Preserve Shop page Page Builder rendering by temporarily setting the global post to the configured Shop page during render.

## Testing
- Confirmed Shop page logs render path and Page Builder content output.
- Confirmed checkout Terms and Conditions path is skipped and no Shop content leakage occurs.
- Ran `php -l siteorigin-panels.php`.

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -530,6 +530,17 @@ class SiteOrigin_Panels {
 			return false;
 		}
 
+		// The shop is a product archive request, even when loop internals shift the queried object.
+		if ( is_post_type_archive( 'product' ) ) {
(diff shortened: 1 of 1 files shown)


AlexGStapleton (approved):
Looks good. There's quite a lot of WC compatibility code in siteorigin-panels.php. It might be worth making a dedicated compat file.

PR #1312, opened 2026-02-26
Move WooCommerce Compatibility Into a Dedicated Compat File

Misplon:
## Summary
- Move WooCommerce-specific Shop content integration out of `SiteOrigin_Panels` and into `compat/woocommerce.php`.
- Load WooCommerce compatibility through `SiteOrigin_Panels_Compatibility`.
- Keep `SiteOrigin_Panels::should_use_woocommerce_shop_page_id()` as a stable delegation point to the new compat class.

## Why
Alex suggested in PR #1310 that WooCommerce compatibility code should live in a dedicated compat file. This change applies that refactor while preserving existing behavior.

## Testing
- `php -l siteorigin-panels.php`
- `php -l inc/compatibility.php`
- `php -l compat/woocommerce.php`

### DIFF
--- a/siteorigin-panels.php
+++ b/siteorigin-panels.php
@@ -75,7 +75,6 @@ class SiteOrigin_Panels {
 
 		// We need to generate fresh post content.
 		add_filter( 'the_content', array( $this, 'generate_post_content' ) );
-		add_filter( 'woocommerce_format_content', array( $this, 'generate_woocommerce_content' ), 10, 2 );
 		add_filter( 'wp_enqueue_scripts', array( $this, 'generate_post_css' ) );
 
 		// Remove the default excerpt function.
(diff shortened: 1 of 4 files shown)


siteorigin-panels.php
```diff
@@ -560,21 +505,7 @@ public function get_post_id() {
 	 * filters while processing non-Shop content.
 	 */
 	public static function should_use_woocommerce_shop_page_id() {
```
AlexGStapleton: We should flag this as [@deprecated](https://docs.phpdoc.org/guide/references/phpdoc/tags/deprecated.html)

AlexGStapleton: Ideally, if we can remove this, we should.

AlexGStapleton: https://github.com/siteorigin/siteorigin-panels/blob/3c032d124c98057c93cf33dac76e40fc36cbb784/inc/renderer.php#L568

https://github.com/siteorigin/siteorigin-panels/blob/3c032d124c98057c93cf33dac76e40fc36cbb784/inc/renderer.php#L824

Should be updated to use `SiteOrigin_Panels_Compat_WooCommerce`

AlexGStapleton (changes requested):
Outside of the `should_use_woocommerce_shop_page_id` related flags, this looks good. Ideally if we can drop the `should_use_woocommerce_shop_page_id` from the main PB Class, we should.

Misplon:
Addressed the remaining review requests in 1439e4e5:

- Removed `SiteOrigin_Panels::should_use_woocommerce_shop_page_id()` from the main class.
- Updated remaining call sites to use `SiteOrigin_Panels_Compat_WooCommerce::should_use_shop_page_id()` directly.
  - `SiteOrigin_Panels::get_post_id()`
  - `SiteOrigin_Panels_Renderer::render()`
  - `SiteOrigin_Panels_Renderer::the_widget()`

This fully removes the `should_use_woocommerce_shop_page_id` flag from the main PB class per review feedback.

AlexGStapleton (approved):
Looks good. Thanks mate

PR #1313, opened 2026-03-04
Implement iframe support for SiteOrigin Layouts Block

AlexGStapleton:
This PR adds API Version 3 support for the SiteOrigin Layouts block, which is required for full compatibility with WordPress 7.0. This PR is very similar to https://github.com/siteorigin/so-widgets-bundle/pull/2276

### High-level overview
- Converted SO Layout Block class to a functional component.
- Moved `editing` state and `BlockControls` into the block `edit()` so the block becomes selectable and the toolbar appears.
- Added concise, iframe-aware logic to resolve whether script/content runs inside an iframe.
- Prevented native `dragstart` events from bubbling out of the panels container (stop `dragstart`) so Gutenberg's HTML5 drag handlers can't hijack interactions.
- Adjusted how the jQuery UI drag helper is inserted in `cell.js`:
  - `appendTo` now uses the dragged element's owner document body (iframe body when relevant), instead of always appending to the admin `body`.
- Patched jQuery UI instances at runtime (after builder render) so their internal `document`/`window` properties point to the iframe's `document`/`window` when the editor is iframed. This fixes:
  - mousemove/mouseup tracking across iframe boundaries;
  - containment and scroll calculations for `draggable`/`sortable` (fixes errors like `overflowOffset is undefined`).

### Testing instructions
To test this PR, follow the following steps:

- Run build:dev and build:release. Test using built release.
- Add a SiteOrigin Layouts block to a page.
- Import the Consultant Prebuilt layout.
- Save.
- Confirm preview looks as expected.
- Reload.
- Confirm preview is present.
- Switch to SO Layout Block to Edit mode.
- Try re-ordering some of the rows and widgets.
- Resize one of the multi-column rows.

Repeat this process in the following setups:
- Current WP release version (pre-WP 7.0).
- 7.0 RC.
- Classic Editor with Page Builder active (any version of WP is fine).

### Notes
Some styling changes are required for the Layouts Block. These aren't directly related to WP 7.0 or this PR, so they'll be included in an upcoming PR.

### DIFF
--- a/compat/js/siteorigin-panels-layout-block.js
+++ b/compat/js/siteorigin-panels-layout-block.js
@@ -131,9 +131,19 @@ function SiteOriginPanelsLayoutBlock(props) {
   }, []); // Setup the panels builder on mount; tear it down on unmount.
 
   wp.element.useEffect(function () {
-    isMountedRef.current = true; // ownerDocument resolves to the iframe document when the editor is iframed (WP 6.3+/7.0).
+    isMountedRef.current = true; // Resolve iframe document and whether script/content run inside an iframe.
 
     var iframeDoc = panelsContainer.current.ownerDocument;
+    var isScriptInIframe = window.self !== window.top;
+    var isContentInIframe = iframeDoc !== window.document;
+    var soDocument = iframeDoc; // Block native HTML5 dragstart so the Block Editor doesn't intercept panel drags.
+
+    var onContainerDragStart = function onContainerDragStart(e) {
+      e.stopPropagation();
+      e.preventDefault();
+    };
+
+    panelsContainer.current.addEventListener('dragstart', onContainerDragStart);
     var $panelsContainer = jQuery(panelsContainer.current);
     var config = {
       editorType: 'standalone',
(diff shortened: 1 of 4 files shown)


Misplon:
Thanks, looks good except for the styling. There is a fair bit to be attended to including field sizes in the WB and button sizes with the new buttons.

Screenshot 2026-03-04 at 22 24 41

Screenshot 2026-03-04 at 22 24 47

Screenshot 2026-03-04 at 22 24 52

Issues that I can see are the font family in the Layout Block. The input height in the Features Widget and perhaps the size of the checkbox input. The new button styling is too close to the edge of the modal.

Misplon:
Prebuilt layout thumbnails are being blocked by CORS. How do you want to handle that?

Misplon:
> Prebuilt layout thumbnails are being blocked by CORS. How do you want to handle that?

## Prebuilt layout thumbnail CORS errors on WP 7.0                                               
WordPress 7.0 introduces `Cross-Origin-Embedder-Policy: credentialless` on block editor pages for the new client-side media processing feature. This  causes the browser to enforce CORS on cross-origin resources, including dynamically-created `` tags like the prebuilt layout thumbnails.           
                                                                                                                                                        
WP 7.0's output buffer (`wp_add_crossorigin_attributes()`) adds `crossorigin="anonymous"` to static HTML elements, but it can't process images created at runtime by JavaScript — which is how the prebuilt layout thumbnails are loaded in `prebuilt.js`.

Since `layouts.siteorigin.com` wasn't returning an `Access-Control-Allow-Origin` header on image responses, the browser blocked them.

**Fix:** Added `Access-Control-Allow-Origin: *` to image responses on `layouts.siteorigin.com` via a Cloudflare Response Header Transform Rule scoped to `layouts.siteorigin.com` image extensions only. These are public layout thumbnails so there's no security concern with the wildcard origin.

Misplon:
Tracking the design/styling follow-up separately in #1315 so this PR can remain focused on the iframe/API v3 compatibility scope.\n\nChecklist issue: https://github.com/siteorigin/siteorigin-panels/issues/1315

AlexGStapleton:
Strangely, I wasn't getting the CORS issue. Really glad you caught that.

PR #1314, opened 2026-03-04
Match Column Count Controls to Integrated Stepper Style

Misplon:
## Summary
- Wrap the row column-count input in an integrated control container.
- Add plus/minus step buttons inside the control on the right.
- Style the control to match the row-width input stepper pattern.
- Hook step-button clicks into existing `changeCellTotal` behavior with `1..12` clamping.

## Validation
- `php -l tpl/js-templates.php`
- Parse check for `js/siteorigin-panels/dialog/row.js`

### DIFF
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -704,7 +704,13 @@ module.exports = panels.view.dialog.extend({
 
 	adjustCellCount: function( e ) {
 		e.preventDefault();
-		var $input = this.$( '.row-set-form input[name="cells"].so-row-field' ).first();
+		var $control = $( e.currentTarget ).closest( '.so-row-count-control' );
+		var $input = $control.find( 'input[name="cells"].so-row-field' ).first();
+
+		if ( ! $input.length ) {
+			$input = this.$( '.row-set-form input[name="cells"].so-row-field' ).first();
+		}
+
 		if ( ! $input.length ) {
 			return;
 		}
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -715,13 +715,38 @@ module.exports = panels.view.dialog.extend({
 			return;
 		}
 
-		var currentValue = parseInt( $input.val(), 10 );
-		if ( _.isNaN( currentValue ) ) {
-			currentValue = this.row && this.row.cells ? this.row.cells.length : 1;
+		var delta = $( e.currentTarget ).hasClass( 'so-row-count-step-up' ) ? 1 : -1;
+		var input = $input.get( 0 );
+		var minValue = parseInt( $input.attr( 'min' ), 10 );
+		var maxValue = parseInt( $input.attr( 'max' ), 10 );
+
+		if ( input && typeof input.stepUp === 'function' && typeof input.stepDown === 'function' ) {
+			if ( delta > 0 ) {
+				input.stepUp();
+			} else {
+				input.stepDown();
+			}
+		} else {
+			var currentValue = parseInt( $input.val(), 10 );
+			if ( _.isNaN( currentValue ) ) {
+				currentValue = this.row && this.row.cells ? this.row.cells.length : 1;
+			}
+			$input.val( currentValue + delta );
 		}
 
-		var delta = $( e.currentTarget ).hasClass( 'so-row-count-step-up' ) ? 1 : -1;
-		var nextValue = Math.max( 1, Math.min( 12, currentValue + delta ) );
+		var nextValue = parseInt( $input.val(), 10 );
+		if ( _.isNaN( nextValue ) ) {
+			nextValue = this.row && this.row.cells ? this.row.cells.length : 1;
+		}
+
+		if ( ! _.isNaN( minValue ) ) {
+			nextValue = Math.max( minValue, nextValue );
+		}
+
+		if ( ! _.isNaN( maxValue ) ) {
+			nextValue = Math.min( maxValue, nextValue );
+		}
+
 		$input.val( nextValue );
 
 		this.changeCellTotal();
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -186,6 +186,8 @@ module.exports = panels.view.dialog.extend({
 			this.$( 'input[name="cells"].so-row-field' ).val( this.model.get( 'cells' ).length );
 		}
 
+		this.$( 'input[name="cells"].so-row-field' ).attr( 'aria-live', 'polite' );
+
 		return this;
 	},
 
--- a/js/siteorigin-panels/dialog/row.js
+++ b/js/siteorigin-panels/dialog/row.js
@@ -719,37 +719,11 @@ module.exports = panels.view.dialog.extend({
 
 		var delta = $( e.currentTarget ).hasClass( 'so-row-count-step-up' ) ? 1 : -1;
 		var input = $input.get( 0 );
-		var minValue = parseInt( $input.attr( 'min' ), 10 );
-		var maxValue = parseInt( $input.attr( 'max' ), 10 );
-
-		if ( input && typeof input.stepUp === 'function' && typeof input.stepDown === 'function' ) {
-			if ( delta > 0 ) {
-				input.stepUp();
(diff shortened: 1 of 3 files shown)


tpl/js-templates.php
```diff
@@ -333,13 +333,20 @@ class="button-primary so-button-mode dashicons so-mode"
...
 		<div class="content">
 
-			<div class="row-set-form">
-				<div class="row-cell-column">
-					<?php
-					esc_html_e( 'Column Count:', 'siteorigin-panels' );
-					echo apply_filters( 'siteorigin_panels_row_column_count_input', '<input type="number" min="1" max="12" name="cells" class="so-row-field" value="2" />' );
-					?>
-				</div>
+				<div class="row-set-form">
```
AlexGStapleton: The wrapper should have r`ole="group"` set, and ideally an `aria-label`. Something like "Column count controls"

AlexGStapleton: Consider adding an invisible span in this div with screen reader instructions. Something like this:

```
<span id="row-cell-description" class="screen-reader-text">
	<?php _e( 'Use arrows keys or type a number between 1 and 12', 'siteorigin-panels' ); ?>
</span>
```

For reference, `screen-reader-text` is a utility class added by WP in the admin and its safe to rely on it existing so no need for additional CSS for this.

tpl/js-templates.php
```diff
@@ -333,13 +333,20 @@ class="button-primary so-button-mode dashicons so-mode"
...
+					<div class="row-cell-column">
+						<?php
+						esc_html_e( 'Column Count:', 'siteorigin-panels' );
```
AlexGStapleton: This should be nested in a label element.

tpl/js-templates.php
```diff
@@ -333,13 +333,20 @@ class="button-primary so-button-mode dashicons so-mode"
...
+						$column_count_input = apply_filters( 'siteorigin_panels_row_column_count_input', '<input type="number" min="1" max="12" name="cells" class="so-row-field" value="2" />' );
```
AlexGStapleton: The input should have `aria-live="polite"` set so the user is informed of how the number changes when there isn't direct focus on it.

tpl/js-templates.php
```diff
@@ -333,13 +333,20 @@ class="button-primary so-button-mode dashicons so-mode"
...
+						?>
+						<div class="so-row-count-control">
+							<?php echo $column_count_input; ?>
+							<div class="so-row-count-stepper">
+								<button type="button" class="so-row-count-step so-row-count-step-up" tabindex="-1">+</button>
+								<button type="button" class="so-row-count-step so-row-count-step-down" tabindex="-1">&minus;</button>
```
AlexGStapleton: The buttons should have tabindex="-1" removed.

js/siteorigin-panels/dialog/row.js
```diff
@@ -701,6 +702,31 @@ module.exports = panels.view.dialog.extend({
...
+			return;
+		}
+
+		var currentValue = parseInt( $input.val(), 10 );
+		if ( _.isNaN( currentValue ) ) {
+			currentValue = this.row && this.row.cells ? this.row.cells.length : 1;
+		}
+
+		var delta = $( e.currentTarget ).hasClass( 'so-row-count-step-up' ) ? 1 : -1;
+		var nextValue = Math.max( 1, Math.min( 12, currentValue + delta ) );
```
AlexGStapleton: Ideally, this amount should be changed using the related HTML API:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepUp

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepDown

This will ensure all related HTML attributes are respected automatically, and ensure future compatibility.

AlexGStapleton (changes requested):
Once the flagged changes are made I think this is good to merge.

Misplon:
Addressed Alex's review feedback in e148ad9:

- Added accessibility semantics to column count controls (`role="group"`, `aria-label`, screen-reader instructions).
- Added explicit label association for the column count input.
- Added `aria-live="polite"` on the default input markup.
- Removed `tabindex="-1"` from stepper buttons.
- Updated stepper logic to use native `stepUp()`/`stepDown()` with bounds-safe fallback handling.

js/siteorigin-panels/dialog/row.js
```diff
@@ -701,6 +704,56 @@ module.exports = panels.view.dialog.extend({
...
+		if ( ! $input.length ) {
+			return;
+		}
+
+		var delta = $( e.currentTarget ).hasClass( 'so-row-count-step-up' ) ? 1 : -1;
+		var input = $input.get( 0 );
+		var minValue = parseInt( $input.attr( 'min' ), 10 );
+		var maxValue = parseInt( $input.attr( 'max' ), 10 );
+
+		if ( input && typeof input.stepUp === 'function' && typeof input.stepDown === 'function' ) {
```
AlexGStapleton: You can safely rely on stepUp/stepDown existing, so no need to check if it does.

https://caniuse.com/mdn-api_htmlinputelement_stepup
https://caniuse.com/mdn-api_htmlinputelement_stepdown

I recommend removing the fallback code and the minValue/maxValue vars, as stepUp/stepDown supports min/max automatically.

AlexGStapleton (approved):
Outside of the flag, this is good to merge.

Misplon:
Addressed the remaining review point in 43b67e5f by simplifying `adjustCellCount()` to rely directly on native `input.stepUp()` / `input.stepDown()` and removing the fallback plus manual min/max clamping logic.

This should fully cover Alex’s feedback in https://github.com/siteorigin/siteorigin-panels/pull/1314#discussion_r2892509808.

Misplon:
@AlexGStapleton Please, could you do a final pass here. Thanks.

tpl/js-templates.php
```diff
@@ -338,9 +338,12 @@ class="button-primary so-button-mode dashicons so-mode"
 							<?php
 							$column_count_input = apply_filters( 'siteorigin_panels_row_column_count_input', '<input type="number" min="1" max="12" name="cells" class="so-row-field" value="2" />' );
 							?>
+							<span class="so-row-count-label"><?php esc_html_e( 'Column Count:', 'siteorigin-panels' ); ?></span>
```
AlexGStapleton: The label should be used here. Also, the label should ideally use [for](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/for).

A label is always better than a span here. If there's a styling issue as a result of using the label, you should correct that rather than moving away from the label.

AlexGStapleton (changes requested):
I'm not too sure why the label was changed to a span.

Misplon:
Addressed the March 8 review comment: the Column Count text is now a proper `<label>` element with `for="so-row-count-input"`, explicitly associated with the number input via a matching `id`. The inner wrapping `<label>` still provides an implicit association as a fallback if the input is replaced via the `siteorigin_panels_row_column_count_input` filter.

All previously addressed accessibility pieces remain intact: `role="group"`, screen-reader instructions, `aria-live="polite"`, focusable step buttons, and native `stepUp()`/`stepDown()`.

@AlexGStapleton ready for another look when you get a chance.

Misplon:
@AlexGStapleton Quick note on PR sequencing/state.

After your March 8 comment on `#1301`, the column-count changes from this PR were also ported into `#1301` so that the main PR would include the same updated stepper/control pattern in both places. `#1301` then received one additional follow-up commit (`16585b37`) for the preview-stepper accessibility/focus-blur behavior.

That means `#1301` now contains the column-count work from this PR as well as its original preview fix.

My suggestion would be to merge `#1301` if the combined result looks right, then close this PR as redundant to avoid duplicating the same column-count changes in two merges.

If you would rather keep the PRs separate, let us know and we can reorganize them accordingly.

AlexGStapleton:
That's definitely a mistake. Only the stepper changes should have been carried over to https://github.com/siteorigin/siteorigin-panels/pull/1301, and only after https://github.com/siteorigin/siteorigin-panels/pull/1314 was merged. Oh well. 

I'm closing https://github.com/siteorigin/siteorigin-panels/pull/1314its entirely in https://github.com/siteorigin/siteorigin-panels/pull/1301 now.

PR #1319, opened 2026-03-18
WP 7: Resolve Inline Saving Issues

AlexGStapleton:
## Summary
- Resolve the WP 7 Block Editor inline row/widget save regression.
- Keep the normal Live Editor save flow on the standard post save path.
- Include the stacked Layout Block inline-save persistence fix from `#1323`.

## Included Work
This PR now includes the original `WP 7: Resolve Inline Saving Issues` change plus the follow-up fix from `#1323`:
- `e4671f2f` `Fix inline save persistence in layout block`
- `f145b2bb` `Live Editor: Use block editor save path`
- merge commit `076e1ba1` bringing `#1323` into this branch

## Root Cause
WP 7 changed the Block Editor save flow enough that the existing inline row/widget save behavior was no longer reliable in the Block Editor path. The follow-up work in `#1323` also addressed Layout Block inline-save persistence and ensured Live Editor uses the correct Block Editor save route when needed.

## Testing
Please verify in WP 7:
- Row inline save persists after saving and immediately refreshing.
- Widget inline save persists after saving and immediately refreshing.
- Layout Block inline save persists.
- Live Editor save follows the correct post-save path.

## Notes
- `#1323` was merged into this branch first, so `#1319` is the PR that should now merge to `develop`.

### DIFF
--- a/compat/js/siteorigin-panels-layout-block.js
+++ b/compat/js/siteorigin-panels-layout-block.js
@@ -198,7 +198,16 @@ function SiteOriginPanelsLayoutBlock(props) {
 
       if (!SiteOriginIsPanelsEqual(initialPanelsData, newPanelsData)) {
         if (typeof onContentChangeRef.current === 'function') {
-          onContentChangeRef.current(newPanelsData);
+          var pendingContentChange = onContentChangeRef.current(newPanelsData);
+
+          if (pendingContentChange && typeof pendingContentChange.then === 'function') {
+            builderViewRef.current.pendingContentChange = pendingContentChange;
+            pendingContentChange["finally"](function () {
+              if (builderViewRef.current && builderViewRef.current.pendingContentChange === pendingContentChange) {
+                builderViewRef.current.pendingContentChange = null;
+              }
+            });
+          }
         }
 
         setLoadingPreview(true);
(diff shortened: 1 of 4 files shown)


Misplon:
I'm not treating this as merge-ready yet.

The main blocker is that `live-editor.js` now routes `closeAndSave()` through `panels.helpers.utils.saveHeartbeat()` instead of triggering the normal WordPress save control. That heartbeat path only repopulates `$_POST["panels_data"]` and calls the plugin's own `save_post()` handler for Page Builder data, so it is not equivalent to a full post save. This looks like it can fix inline row/widget saving while regressing the Live Editor save button.

I haven't run the manual `build:dev` / full-build runtime test yet, but I'd want that path adjusted and then verified in WP 7 before merging. There's also a follow-up integration check worth doing here: the new block-editor save path calls `core/editor.savePost()`, while the plugin still injects Page Builder data via Heartbeat, so that combination should be confirmed explicitly during testing.

Misplon:
In addition to the above, the visible behavior in the Classic Editor seems ok but not in the Block Editor.

AlexGStapleton:
> That heartbeat path only repopulates $_POST["panels_data"] and calls the plugin's own save_post() handler for Page Builder data, so it is not equivalent to a full post save. This looks like it can fix inline row/widget saving while regressing the Live Editor save button.

Ah. You're right. I'll change that shortly.

Personally, I don't think the Live Editor save button should save the entire page (that's well out of the scope of that button), but if we were to change that, it would be better to handle that in a dedicated focused PR rather than this PR.

> There's also a follow-up integration check worth doing here: the new block-editor save path calls core/editor.savePost(), while the plugin still injects Page Builder data via Heartbeat, so that combination should be confirmed explicitly during testing.

That's intended as a fallback, as this change is only a requirement for WP 7 plus while the old version will work with all supported prior versions.

Misplon:
Pushed a follow-up fix in 2201f3e4.

This restores the Live Editor save button to the normal post save control instead of routing `closeAndSave()` through `saveHeartbeat()`. The inline row/widget save path in `utils.js` is left intact for the WP 7 block-editor fix.

That keeps the branch scoped to the inline saving issue while removing the Live Editor regression blocker.

Misplon:
Pushed another follow-up in 66d2e3dd.

The block editor inline row/widget save path no longer depends on heartbeat injection. Row and widget inline saves now use the editor save flow directly in the Block Editor, while Classic Editor keeps using the existing heartbeat path.

That should address the case where `savePost()` completed but the Page Builder data only existed on the heartbeat bridge.

Misplon:
In it's current form, I don't believe this branch is functional. If you make a row level change, say setting a row ID, save and wait for the row save to complete and then immediately refresh the page, the change is lost. When you exit the block, the page auto-saves so in that sequence of events it's easy to think that inline saving is working but it's not actually, it's relying on the page auto saving.

Misplon:
I reverted my changes on this branch and reset it back to 0ba555e9, so #1319 is back to Alex's original WP 7 inline-saving changes only.

AlexGStapleton:
I've reviewed https://github.com/siteorigin/siteorigin-panels/pull/1323 and flagged one issue. Based on the existence of that PR, I don't think your other flags in this PR are applicable anymore, but let me know if they are and I'll action them asap.

PR #1320, opened 2026-03-19
Refine Layout Block top toolbar mode control

Misplon:
## Summary
- switch the Layout Block mode control from Toolbar to ToolbarGroup for correct top toolbar rendering
- remove the extra manual toolbar button classes from the mode control button
- reserve 2px inside the Layout Block container so Gutenberg's selected ring does not overlap the block UI in topbar mode

## Testing
- verified locally in the block editor with the toolbar set to topbar
- confirmed the temporary local changes were removed from the WP 7 styling branch after porting them here


Misplon:
Closing this because the proposed change does not address the reported top toolbar focus artifact. Reworking from scratch on a new branch.

PR #1322, opened 2026-03-19
Fix Layout Block Context Menu in the Block Editor Iframe

Misplon:
## Summary
- Keep the context menu activation listeners bound to the Block Editor iframe document where the Layout Block builder lives.
- Render the visible context menu in the top admin document and translate iframe click coordinates into top-window coordinates before opening it.
- Tear down builder and menu instances on Layout Block unmount so stale listeners do not survive editor remounts.

## Root Cause
This turned out to be two separate problems in the iframe editor path:

1. Stale builder/menu instances were surviving Layout Block remounts after refresh, leaving old document-level listeners behind.
2. Even once the active menu instance was correct, rendering the visible menu inside the iframe document was unreliable after reload. The menu activated and populated, but its rendered box collapsed instead of displaying.

The final fix keeps right-click detection in the iframe, but renders the actual menu in the top admin document where the menu styles and layout behave consistently.

## Testing
- `node --check js/siteorigin-panels/utils/menu.js`
- `node --check js/siteorigin-panels/view/builder.js`
- `node --check compat/js/siteorigin-panels-layout-block.js`
- `node --check js/siteorigin-panels.js`
- Rebuilt `compat/js/siteorigin-panels-layout-block.js` with Babel and `js/siteorigin-panels.js` with Browserify locally
- Current manual browser checks look good, but more browser verification is still useful before merge

## Notes
- `gulp` build tasks are blocked locally by the repo's legacy `node-sass` dependency under Node `v24.7.0`, so the JS assets were rebuilt directly with the installed Babel/Browserify tooling instead.

### DIFF
--- a/js/siteorigin-panels/utils/menu.js
+++ b/js/siteorigin-panels/utils/menu.js
@@ -35,14 +35,111 @@ module.exports = Backbone.View.extend( {
 		return this.contextWindow || contextDocument.defaultView || window;
 	},
 
-	getOutsideClickWindows: function () {
-		var interactionWindows = [ this.getContextWindow() ];
+	getRenderWindow: function () {
+		var renderWindow = this.getContextWindow();
 
-		if ( window !== interactionWindows[ 0 ] ) {
-			interactionWindows.push( window );
+		try {
+			if ( renderWindow.top && renderWindow.top.document ) {
+				renderWindow = renderWindow.top;
+			}
+		}
+		catch ( err ) {}
+
+		return renderWindow;
+	},
+
+	getRenderDocument: function () {
+		var renderWindow = this.getRenderWindow();
+		return renderWindow.document || document;
+	},
+
+	getUniqueWindows: function( windows ) {
+		var uniqueWindows = [];
+
+		_.each( windows, function( candidateWindow ) {
+			if ( candidateWindow && uniqueWindows.indexOf( candidateWindow ) === -1 ) {
+				uniqueWindows.push( candidateWindow );
+			}
+		} );
+
+		return uniqueWindows;
+	},
+
+	getOutsideClickWindows: function () {
+		return this.getUniqueWindows( [
+			this.getContextWindow(),
+			this.getRenderWindow(),
+			window
+		] );
+	},
+
+	getKeyupWindows: function () {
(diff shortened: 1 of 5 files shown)


Misplon:
@AlexGStapleton The iframe compatibility merge broke the contextual menu in the Block Editor. That issue is in develop. This PR is to fix the issue.

js/siteorigin-panels/utils/menu.js
```diff
@@ -15,18 +15,82 @@ module.exports = Backbone.View.extend( {
...
+	getContextWindow: function () {
+		var contextDocument = this.getContextDocument();
+		return this.contextWindow || contextDocument.defaultView || window;
+	},
+
+	getInteractionWindows: function () {
+		var interactionWindows = [ this.getContextWindow() ];
+
+		if ( window !== interactionWindows[ 0 ] ) {
+			interactionWindows.push( window );
```
AlexGStapleton: This isn't ideal. Ideally, the events should only be set up to target the one context, not all of them. Supporting both could result in unintended consequences.

Based on the other changes, this was done to enable click detection outside the Layouts Block. That specific area should have this check (`window !== interactionWindows[ 0 ]`) rather than all instances of `this.getInteractionWindows()`

AlexGStapleton (changes requested):
Outside of the one flag, this is good to go.

js/siteorigin-panels/utils/menu.js
Misplon: Updated in 86c69fd8. I narrowed the multi-window behavior so only outside-click detection spans both the iframe window and parent admin window. The context-menu listener and keyup handling stay scoped to the active builder context now.

Misplon:
Additional context for review: the final fix ended up being narrower than the original PR description. We confirmed this was not just a wrong-document listener issue. Refresh was also leaking stale builder/menu instances, and after that was fixed the remaining failure was that the active menu would populate but not render correctly when attached inside the iframe document after reload. That is why the current implementation keeps the interaction listener in the iframe, but renders the visible menu in the top admin document and translates coordinates across the iframe boundary.

AlexGStapleton (approved):

PR #1323, opened 2026-03-24
Fix layout block inline save persistence

Misplon:
## Summary
This PR is stacked on `wp-7-resolve-inline-saving-block-editor`.

The original WP 7 inline-saving work identified a real problem, but the failing block-editor repro turned out not to be the old metabox/heartbeat save path. In the Layout Block flow, inline save was racing the block's async `panelsData` sanitization/update cycle, so WordPress could save stale block attributes and lose the change on hard refresh.

This patch fixes that by:
- tracking pending Layout Block content updates while `panelsData` is being sanitized and written back with `setAttributes()`
- making inline save wait for that pending update in standalone block-editor builders
- waiting for `core/editor` post-saving locks to clear before dispatching `savePost()`
- waiting for the Gutenberg save to complete before re-enabling the inline-save button
- keeping the safer Live Editor behavior from the earlier investigation instead of routing it through the inline save helper

## Findings / route taken
- The deep-research pass was useful in one respect: it correctly highlighted that the original PR mixed Heartbeat-driven inline save semantics with Block Editor post-save semantics.
- That led to the first hypothesis that the failure was the old Heartbeat transport being replaced too broadly.
- Your repro narrowed it further: Layout Block row/widget edits were still not persisting after hard refresh even when the UI reported save completion.
- Tracing the Layout Block integration showed the real issue was ordering. Inline save could run before the async `so_panels_builder_content_json` sanitization round-trip had finished updating block attributes.
- The effective fix therefore had to land in the Layout Block attribute pipeline and in the inline-save helper's block-editor save timing, not just in the legacy Heartbeat path.

## Testing checklist
- [x] In the block editor, insert a `SiteOrigin Layout` block, add a row, set a row ID, click `Save Now`, hard refresh, and confirm the row ID persists.
- [x] In the block editor, insert a `SiteOrigin Layout` block, add an Editor widget, use the normal save once, then edit widget text, click `Save Now`, hard refresh, and confirm the edited text persists.
- [x] Repeat the same checks with an existing Layout Block instead of a newly inserted one.
- [x] Confirm inline save re-enables after completion and does not get stuck disabled.
- [x] Confirm no regression in classic Page Builder row/widget inline save flows.
- [x] Confirm Live Editor save still follows the normal post-save path.

### DIFF
--- a/js/siteorigin-panels/view/live-editor.js
+++ b/js/siteorigin-panels/view/live-editor.js
@@ -124,16 +124,18 @@ module.exports = Backbone.View.extend( {
 			// The live editor requires a saved draft post, so we'll create one for auto-draft posts
 			var thisView = this;
 
-			// Set a temporary post title so the autosave triggers properly
-			if( $('#title[name="post_title"]' ).val() === '' ) {
-				$('#title[name="post_title"]' ).val( panelsOptions.loc.draft ).trigger('keydown');
-			}
+			if ( wp.autosave ) {
+				// Set a temporary post title so the autosave triggers properly
+				if( $('#title[name="post_title"]' ).val() === '' ) {
+					$('#title[name="post_title"]' ).val( panelsOptions.loc.draft ).trigger('keydown');
+				}
 
-			$( document ).one( 'heartbeat-tick.autosave', function(){
-				thisView.autoSaved = true;
-				thisView.refreshPreview( thisView.builder.model.getPanelsData() );
-			} );
-			panels.helpers.utils.saveHeartbeat( thisView );
+				$( document ).one( 'heartbeat-tick.autosave', function(){
+					thisView.autoSaved = true;
+					thisView.refreshPreview( thisView.builder.model.getPanelsData() );
+				} );
+				wp.autosave.server.triggerSave();
+			}
 		}
 	},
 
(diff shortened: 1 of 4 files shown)


js/siteorigin-panels/view/live-editor.js
```diff
@@ -167,7 +169,8 @@ module.exports = Backbone.View.extend( {
 	 */
 	closeAndSave: function(){
 		this.close( false );
-		panels.helpers.utils.saveHeartbeat( this );
+		// Finds the submit input for saving without publishing draft posts.
+		$( '#submitdiv input[type="submit"][name="save"], .editor-post-publish-button, .edit-widgets-header__actions .is-primary' )[0].click();
```
AlexGStapleton: These selectors don't exist in the Block Editor. Consider doing a shouldUseBlockEditorSave check, and then using saveBlockEditor or this click, depending upon the result.

AlexGStapleton (changes requested):
Outside of the one flag, this looks good to go. There are a few things that could be changed, but due to timing, I'm doing a high level "only flag if it could break" type of review. If you would prefer a full review, please let me know.

I noticed no issues or regressions in the Classic Editor, or other test setups.

js/siteorigin-panels/view/live-editor.js
Misplon: Updated in f145b2bb. `closeAndSave()` now checks `shouldUseBlockEditorSave()` and calls `saveBlockEditor()` in block editor contexts, while keeping the existing click-through save path everywhere else.

PR #1334, opened 2026-05-11
Restore post globals after admin renders

Misplon:
## Summary

- Wrap admin-side Page Builder render calls so WordPress post globals are restored after rendering.
- Apply the wrapper to generated post content, classic/admin previews, live editor preview, layout block AJAX previews, and layout block editor/save validation renders.
- Keep the wrapper signature compatible with `SiteOrigin_Panels_Renderer::render()` by passing `$layout_data` by reference.
- Preserve the renderer's own `$siteorigin_panels_current_post` state as well as WordPress post globals.

## Root Cause

Some widgets can run their own post queries during admin preview/content renders and leave globals such as `$post` pointing at a queried item. Those leaked globals can affect later admin rendering, including permalink-dependent widgets such as carousels.

The same issue can affect later output in admin preview responses. For example, live editor preview renders before `wp_footer()`, and layout block editor/save validation can render Page Builder content outside the original `inc/admin.php` wrapper call sites.

## Follow-up Review Update

After additional review, commit `87659594` expanded the fix to cover the remaining admin/editor render surfaces:

- Made `SiteOrigin_Panels_Admin::render_and_restore_post_globals()` public/static so preview paths outside `inc/admin.php` can use the same wrapper.
- Added `siteorigin_panels_current_post` to the snapshot/restore list to avoid stale renderer state after early returns or failed renders.
- Routed `tpl/live-editor-preview.php` rendering through the wrapper so leaked post globals cannot affect later footer output in the AJAX preview document.
- Routed Layout Block editor and server-side validation renders through the wrapper while leaving normal frontend Layout Block rendering direct.

## Validation

- `php -l inc/admin.php`
- `php -l compat/layout-block.php`
- `php -l tpl/live-editor-preview.php`
- `git diff --check`
- Smoke test confirmed normal restore, exception restore, `siteorigin_panels_current_post` restoration, and by-reference `$layout_data` propagation.
- Manual retest by Andrew confirmed the original warning fix looked okay before publishing.

## Manual Tests

- Corp demo Sandbox page: add only Cards Carousel, save, confirm page slug/permalink and DB `post_name` remain the page's slug.
- Classic editor/Page Builder preview and builder content AJAX with Cards Carousel.
- Block editor Layout Block AJAX preview with Cards Carousel.
- Layout Block save/server-side validation with Cards Carousel.
- Live editor preview with Cards Carousel, checking for footer/preview oddities.

### DIFF
--- a/compat/layout-block.php
+++ b/compat/layout-block.php
@@ -151,7 +151,11 @@ class SiteOrigin_Panels_Compat_Layout_Block {
 			SiteOrigin_Panels_Post_Content_Filters::add_filters( true );
 		}
 
-		$rendered_layout = SiteOrigin_Panels::renderer()->render( $builder_id, ! $is_editing, $panels_data );
+		if ( $is_editing || ! $this->return_layout ) {
+			$rendered_layout = SiteOrigin_Panels_Admin::render_and_restore_post_globals( $builder_id, ! $is_editing, $panels_data );
+		} else {
+			$rendered_layout = SiteOrigin_Panels::renderer()->render( $builder_id, true, $panels_data );
+		}
 
 		if ( $is_editing ) {
 			SiteOrigin_Panels_Post_Content_Filters::remove_filters( true );
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -260,7 +260,7 @@ class SiteOrigin_Panels_Admin {
 
 				SiteOrigin_Panels_Post_Content_Filters::add_filters();
 				$GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] = true;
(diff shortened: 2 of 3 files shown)


Misplon:
@AlexGStapleton The issue being resolved is as follows:

- Using the current release version of PB, on the Corp demo add only a Cards Carousel Widget to the Sandbox page: https://demo.siteorigin.com/corp/wp-admin/post.php?post=6051&action=edit.
- Save.
- Note that the the page permalink adopts the slug of the first post in the carousel.

I recall submitting a gate related SiteOrigin PR to resolve this issue in the past. I can't find that PR at the moment but the change seemed like a valid tightening regardless. That said, it wasn't the root cause of the issue.

AlexGStapleton (approved):
Looks good on my end.

PR #1346, opened 2026-07-13
Guard Layout Block sanitization against non-array panels data

### LINKED ISSUE #1342: Layout Block: guard sanitize path against non-array panelsData attribute (PHP 8 fatal)
## Summary
`sanitize_block()` (`compat/layout-block.php`) checks `empty( $block['attrs']['panelsData'] )` but never confirms `panelsData` is an array. A block comment carrying a non-array attribute value — e.g. `<!-- wp:siteorigin-panels/layout-block {"panelsData":"x"} /-->` — passes that check and reaches `sanitize_panels_data()`, whose first statement is `unset( $panels_data['sanitize_signature'] );`. On PHP 8 that fatals with **"Cannot unset string offsets"**.

## Impact
- The fatal occurs during the save request that carries the malformed content; the request fails, no data loss or further effect.
- Reachable from all five save-validation surfaces that funnel through `sanitize_block()`: `rest_pre_insert_{post_type}`, `rest_pre_insert_wp_block`, `pre_update_option_widget_block`, and `wp_insert_post_data`.
- Pre-existing since the original REST validation path; the newer surfaces extended reachability but did not change severity.
- `verify_panels_data()` itself is unaffected (`empty()` on an illegal string offset is false-safe).

## Suggested fix
A one-line `is_array( $panels_data )` guard — either in `sanitize_block()`'s early-return condition or at the top of `sanitize_panels_data()` — closes this for all five surfaces at once. Should ship with a regression test following the existing `tests/` conventions (self-contained, Brain Monkey, no anonymous classes).

Ref: flagged in the review of the save-surface-hardening merge (`3bf95ecf`).

oiahoon:
## Summary
- return malformed non-array `panelsData` values before the shared save-time sanitizer
- add a regression covering the `wp_insert_post_data` validation path

Closes #1342

## Testing
- PHPUnit suite: 25 tests, 78 assertions
- `php -l compat/layout-block.php`
- `php -l tests/LayoutBlockInsertPostDataValidationTest.php`
- `git diff --check`

### DIFF
--- a/compat/layout-block.php
+++ b/compat/layout-block.php
@@ -648,7 +648,8 @@ class SiteOrigin_Panels_Compat_Layout_Block {
 	public function sanitize_block( $block ) {
 		if (
 			empty( $block['attrs'] ) ||
-			empty( $block['attrs']['panelsData'] )
+			empty( $block['attrs']['panelsData'] ) ||
+			! is_array( $block['attrs']['panelsData'] )
 		) {
 			return $block;
 		}
--- a/tests/LayoutBlockInsertPostDataValidationTest.php
+++ b/tests/LayoutBlockInsertPostDataValidationTest.php
@@ -79,6 +79,8 @@ class InsertPostDataRendererStub {
  *     with panelsData intact and verifying — the case that catches the
  *     unslash/reslash bug found in plan review (case (d) alone would pass
  *     with that bug present).
+ * (f) A malformed non-array panelsData value passes through unchanged instead
+ *     of reaching array-only sanitization and causing a PHP fatal error.
  *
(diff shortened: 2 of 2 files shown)


Misplon:
Thanks so much for this, @oiahoon — spot-on root cause and a clean regression test.

We added two small follow-up commits on top of yours: the same `is_array()` check also needs to sit at the `render_layout_block()` entry, since a non-array `panelsData` hits the identical fatal on the front-end render path (via `prepare_render_panels_data()` → `sanitize_panels_data()`), plus a belt-and-braces guard in `sanitize_panels_data()` itself. Added a couple of render-path tests to cover it. Your save-path guard and test are kept as-is.

Really appreciate the contribution.

PR #1367, opened 2026-08-12
Fix builder rendering failures when stored widget or row style is null (#1366)

### LINKED ISSUE #1366: Builder Rendering Fails When a Stored Widget or Row Style Is Null
**Symptom**

If a page's stored panels_data contains `"style": null` on a widget or row, the builder throws while rendering:

```
Cannot read properties of null (reading 'disable_widget')
```

The exception halts the render loop, so every widget row on the page displays as a blank box with no title, and none of them open. The builder never finishes loading, so the page cannot be re-saved and the bad value never self-heals.

**Root Cause**

Three gaps cooperate:

1. `js/siteorigin-panels/view/widget.js` — `toggleVisibilityFade()` guards only `typeof styles == 'undefined'`. `typeof null` is `'object'`, so a null passes the guard and `styles['disable_widget']` throws.
2. `js/siteorigin-panels/model/builder.js:133` — the load path checks `!_.isUndefined( panels_info.style )`, so a stored null overwrites the model's `{}` default.
3. `inc/styles-admin.php:589` — sanitization skips empty styles instead of normalizing them, so a null survives every save.

Panels never writes `"style": null` itself. Legacy data, or third-party code that filters panels_data, can leave the key set to null.

**Proposed Hardening**

- `toggleVisibilityFade()`: return early unless `styles` is an object.
- `model/builder.js`: only `set( 'style', ... )` when the value is an object.
- `sanitize_all()`: remove style keys that are null instead of skipping them.
- Audit the row and cell views for the same `typeof`-only guard.

AlexGStapleton:
Resolve https://github.com/siteorigin/siteorigin-panels/issues/1366

- Frontend View Guards (widget.js, row.js): Prevent runtime exceptions in toggleVisibilityFade() when encountering null or non-object style values. (bea0fba2c3477bc89b52fced75305c6efa9e1c63)
- Model Loading (builder.js, cell.js, row.js, widget.js): Guard against assigning stored null styles and normalise uninitialized or invalid styles to {} during model instantiation. (08469963270042dde921f2b052d7a98d7c426877)
- Backend Sanitisation (styles-admin.php): Unset/remove null style entries during sanitize_all() so that corrupted style values do not survive save operations. (6b3d3341a88d3463f327f7b4144a62d2d01a0869)

### DIFF
--- a/inc/styles-admin.php
+++ b/inc/styles-admin.php
@@ -586,11 +586,15 @@ class SiteOrigin_Panels_Styles_Admin {
 					);
 				}
 
-				if ( empty( $panels_data['widgets'][ $i ]['panels_info']['style'] ) ) {
-					continue;
+				if ( isset( $panels_data['widgets'][ $i ]['panels_info']['style'] ) ) {
+					if ( is_null( $panels_data['widgets'][ $i ]['panels_info']['style'] ) || ! is_array( $panels_data['widgets'][ $i ]['panels_info']['style'] ) ) {
+						unset( $panels_data['widgets'][ $i ]['panels_info']['style'] );
+					} elseif ( empty( $panels_data['widgets'][ $i ]['panels_info']['style'] ) ) {
+						continue;
+					} else {
+						$panels_data['widgets'][ $i ]['panels_info']['style'] = $this->sanitize_style_fields( 'widget', $panels_data['widgets'][ $i ]['panels_info']['style'] );
+					}
 				}
-
-				$panels_data['widgets'][ $i ]['panels_info']['style'] = $this->sanitize_style_fields( 'widget', $panels_data['widgets'][ $i ]['panels_info']['style'] );
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


Misplon:
Reviewed this while deciding whether it lets us close siteorigin/siteorigin-premium#1293. The JS half looks right to me: `toggleVisibilityFade()` now returns for `null` and non-objects, which covers the read in `checkIfStyleExists()` at `js/siteorigin-panels/view/widget.js:151` where the original `Cannot read properties of null (reading 'disable_widget')` came from, and the model normalisation stops a stored null reaching the view in the first place.

Two things about the `sanitize_all()` half, though.

**1. The null branch is unreachable.** `inc/styles-admin.php`, all three blocks:

```php
if ( isset( $panels_data['widgets'][ $i ]['panels_info']['style'] ) ) {
    if ( is_null( $panels_data['widgets'][ $i ]['panels_info']['style'] ) || ! is_array( ... ) ) {
        unset( ... );
```

`isset()` is `false` for a key that exists with a `null` value, so for an actual `null` the outer condition never opens and the `is_null()` check never runs:

```
$d = array( 'widgets' => array( array( 'panels_info' => array( 'style' => null ) ) ) );
isset( $d['widgets'][0]['panels_info']['style'] )            // false
array_key_exists( 'style', $d['widgets'][0]['panels_info'] ) // true
```

Running the block against that input leaves `'style' => NULL` in place. The `! is_array()` half still works for a non-null scalar, so the branch isn't dead for every input — just for the one the PR is about. `array_key_exists()` in place of `isset()` fixes it.

**2. Ordering means this hook can't catch the null we actually hit.** Page Builder sanitises before it applies the pre-save filter:

```php
$panels_data = SiteOrigin_Panels_Styles_Admin::single()->sanitize_all( $panels_data );
$panels_data = apply_filters( 'siteorigin_panels_data_pre_save', $panels_data, $post, $post_id );
```

`inc/admin.php:290-291`, and the same order at `inc/admin.php:1011` and `inc/abilities.php:412`. Toggle Visibility writes the null from `pb_migrate_settings_save()`, which is on that later filter, so it lands after `sanitize_all()` has already run. Even with fix 1 applied, this path cannot clean it up.

---

**What I'm asking for:**

- Change `isset()` to `array_key_exists()` in the three blocks, so the null cleanup does what the PR description says it does.
- Consider whether `sanitize_all()` is the right place at all given the ordering above, or whether that cleanup wants to be after the pre-save filter.

**Why, and what it means for premium#1293:** with only this PR applied, the builder stops crashing, but Toggle Visibility keeps writing `style => null` on every save. The corruption carries on accumulating, just invisibly. That's why we're keeping premium#1293 open rather than closing it in favour of this: it stops the write at the source and repairs data already stored. The two look complementary to us — this one makes Page Builder resilient to bad data from anywhere, #1293 stops Premium producing it.

Not asking you to change that judgement, just explaining why #1293 isn't being closed. Happy to be told I've read the ordering wrong.

Misplon (dismissed):
Thanks, mate. Please, see my requested changes above.

Misplon:
The two changes requested above are the first two commits of #1374, which is now stacked on this branch: `bbd27ce3` (one helper using `array_key_exists()`, run inside `sanitize_all()` and again after `siteorigin_panels_data_pre_save` at all three save sites) and `14960a26` (tests, including the real `save_post()` with a null-writing pre-save callback). Nothing on this branch was touched. Suggested order: merge this PR as-is, then #1374. If you'd rather carry the two commits here, cherry-pick them.

PR #1372, opened 2026-09-02
Copy Content: keep shortcodes intact in the post content mirror

### LINKED ISSUE #1370: Copy Content Bakes Shortcode Output Into the Post Content Mirror
## What's wrong

Copy Content is on by default. It renders the layout at save time and stores the result in `post_content`. Widget shortcodes are executed during that render, so their output is baked into the stored copy rather than the shortcode itself being kept.

A form shortcode, for example, ends up stored as its rendered placeholder markup. Wherever the mirror is displayed instead of a live Page Builder render, the visitor gets a dead form and the plugin's JS and CSS are never enqueued, because the plugin enqueues them at shortcode execution time. That happens when `the_content` takeover is disabled via `siteorigin_panels_filter_content_enabled`, in feeds, in search indexing, when Page Builder is deactivated, and when kses strips the inline script on save.

Baked output also defeats any plugin that scans `post_content` with `has_shortcode()` to decide whether to enqueue its front end assets.

Related to #362, which asked for shortcodes to be kept intact in Copy Content.

## The change

In `inc/post-content-filters.php`, empty the global `$shortcode_tags` registry while the mirror renders, and restore it afterwards. Core's `do_shortcode()` short-circuits on an empty registry and returns its input unchanged, so every widget keeps raw shortcodes in the copied content, from any widget and any plugin, with no per-widget compatibility work. The raw shortcodes then run normally through `the_content` when the mirror itself is displayed.

Details that matter:

- Suspension is depth counted, so nested scopes stay suspended until the outermost one closes. Nested scopes are counted without re-consulting the opt-out filter, so a mid-render change in its result cannot desync the accounting.
- Shortcodes registered lazily during the render survive the restore; original registrations win on conflict.
- Excess remove calls are no-ops.
- Sites relying on baked output can opt out through a new `siteorigin_panels_post_content_keep_shortcodes` filter.
- `inc/admin.php` wraps the three post content render sites in `try`/`finally`, so a widget that throws mid-render cannot leave the filters active or the registry suspended for the rest of the request.
- `inc/widget-shortcode.php` adds `WP_Widget_Custom_HTML` to the shortcode widget allow list.

## Status

A patch exists and has been verified to apply cleanly to `develop`. The change was previously built, reviewed and shipped to a reporter as a beta, and verified end to end on a live install: the form rendered with its assets in the Editor widget, in the Custom HTML widget, and when the copy content mirror itself was displayed, with AJAX submission working and Block Editor paths untouched.

The branch it was built on was never pushed and no longer exists, so it needs re-applying. The tests that went with it were not recovered and need rewriting: `tests/copy-content/CopyContentParityTest.php` on `develop` has no shortcode coverage.

Companion change in the Widgets Bundle, needed for the case where an older Page Builder is paired with a newer Widgets Bundle.

Misplon:
Fixes #1370.

Suspends shortcode execution during the post content mirror render: the registry is emptied for the render's duration and a `pre_do_shortcode_tag` guard short-circuits shortcodes registered mid-render. Raw shortcodes survive in `post_content`, so `has_shortcode()` scans work and the shortcode renders live wherever the mirror is displayed. Restore of the fix shipped to a reporter as 2.36.1-beta.1.

- Depth-counted nested scopes; an opted-out outer render is inherited by nested scopes.
- `siteorigin_panels_post_content_keep_shortcodes` filter opts back into baked output for widgets whose shortcodes run through the suspended registry, such as the core Text widget. The Widgets Bundle Editor widget carries its own render guard and does not yet consult this filter; tracked in siteorigin/so-widgets-bundle#2365.
- try/finally at all three render sites restores state if a render throws; the opt-out consult runs before any registration so a throwing callback leaves nothing behind.
- `WP_Widget_Custom_HTML` added to the text-widget allow-list.

Tests: 14-method `PostContentShortcodeSuspensionTest` plus a `CopyContentParityTest` exception-safety case, red against develop, green here. All five suites pass.

Companion PR: siteorigin/so-widgets-bundle covering the Editor widget for older-Panels installs.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -359,10 +359,17 @@ class SiteOrigin_Panels_Admin {
 
 			SiteOrigin_Panels_Post_Content_Filters::add_filters();
 			$GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] = true;
-			$post_content = self::render_and_restore_post_globals( $layout_id, false, $panels_data );
-			$post_css = SiteOrigin_Panels::renderer()->generate_css( $layout_id, $panels_data );
-			SiteOrigin_Panels_Post_Content_Filters::remove_filters();
-			unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
+
+			try {
+				$post_content = self::render_and_restore_post_globals( $layout_id, false, $panels_data );
+				$post_css = SiteOrigin_Panels::renderer()->generate_css( $layout_id, $panels_data );
+			} finally {
+				// A widget that throws mid-render must not leave the content
+				// render filters active or the shortcode registry suspended
+				// for the rest of the request.
+				SiteOrigin_Panels_Post_Content_Filters::remove_filters();
+				unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
+			}
 
 			// Update the post_content.
 			$post->post_content = $post_content;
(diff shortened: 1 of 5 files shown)


Misplon:
Reverted on `develop` (8dbbf5ba) before release. Re-landed for review as #1373 with the full context there.

Issue #362, opened 2017-03-20
Copy Copy: Add ability to keep shortcodes intact when copied


AlexGStapleton:
Follow up to #361. We should add the ability to keep shortcodes intact when copied rather than running them which can lead to a whole lot of undesired data appearing in the post. This normally is okay but it can prove to be problematic with plugins Yoast if invalid data is analyzed.

gregpriday:
Yeah, mostly the same implementation as #361. Resolved by https://github.com/siteorigin/siteorigin-panels/commit/d181a2b455e29be391058343233138c8f72d3dad

AlexGStapleton:
In hindsight, I was overthinking things. :/

Issue #361, opened 2017-03-20
Copy Content: Add the ability to filter what's copied


AlexGStapleton:
In certain instances, user's (or widgets) may not wish to have certain things included in the copied content.  Things the SiteOrigin Slider which will add the pagination numbers to the copied content. Another example is, say, a slider like Meta Slider. It adds a bunch of code that users will never want to be included in the excerpt or analyzed by plugins like Yoast.

gregpriday:
I added a really basic function that other widgets can check to see if the content being rendered is for the database (in which case it shouldn't be processed). We could add more support to the Widgets Bundle to produce shortcodes instead of code for all non-text based widgets.

PR #1373, opened 2026-09-04
Copy Content: Keep Shortcodes Intact in the Post Content Mirror

### LINKED ISSUE #1370: Copy Content Bakes Shortcode Output Into the Post Content Mirror

Misplon:
Re-lands #1372, which was merged and then reverted on `develop` (8dbbf5ba) so it could get a proper review before release. Fixes #1370.

**What it changes.** During the Copy Content render, Page Builder empties the shortcode registry, so every widget from every plugin keeps raw `[shortcode]` text in the `post_content` copy instead of baked output. Core Text widget shortcodes have baked into the mirror since 2017. This reverses that for everyone, with `siteorigin_panels_post_content_keep_shortcodes` as the opt-out.

**Why it exists.** A form shortcode baked into the mirror displayed as a dead form wherever the mirror was shown, and the form plugin's `has_shortcode()` check never found its shortcode, so its assets didn't load.

**The history.** @AlexGStapleton asked for shortcodes to stay intact in the copy in #361 and #362. Greg's answer was a per-widget flag, which the Editor widget used for seven years. so-widgets-bundle@16abcb6a (1.58.11, March 2024) moved `do_shortcode` outside that guard for the Block Editor, and the Editor widget started baking. so-widgets-bundle#2364, already on develop, puts it back and covers the reporter's case on its own.

**The question for review.** Is the global version worth the surface area? The risk is anything reading `post_content` raw and expecting HTML: headless or REST consumers, a search indexer with shortcode expansion off, an email plugin. Front-end rendering is untouched.

Tests: 14-method `PostContentShortcodeSuspensionTest` plus a `CopyContentParityTest` exception-safety case. All five suites pass. Verified live on a local site through the classic editor save path.

Companion: so-widgets-bundle#2366 (draft) makes the Editor widget honour the opt-out filter. Merge it only if this lands.

### DIFF
--- a/inc/admin.php
+++ b/inc/admin.php
@@ -359,10 +359,17 @@ class SiteOrigin_Panels_Admin {
 
 			SiteOrigin_Panels_Post_Content_Filters::add_filters();
 			$GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] = true;
-			$post_content = self::render_and_restore_post_globals( $layout_id, false, $panels_data );
-			$post_css = SiteOrigin_Panels::renderer()->generate_css( $layout_id, $panels_data );
-			SiteOrigin_Panels_Post_Content_Filters::remove_filters();
-			unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
+
+			try {
+				$post_content = self::render_and_restore_post_globals( $layout_id, false, $panels_data );
+				$post_css = SiteOrigin_Panels::renderer()->generate_css( $layout_id, $panels_data );
+			} finally {
+				// A widget that throws mid-render must not leave the content
+				// render filters active or the shortcode registry suspended
+				// for the rest of the request.
+				SiteOrigin_Panels_Post_Content_Filters::remove_filters();
+				unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
+			}
 
 			// Update the post_content.
 			$post->post_content = $post_content;
@@ -1604,9 +1611,13 @@ class SiteOrigin_Panels_Admin {
 		// Create a version of the builder data for post content.
 		SiteOrigin_Panels_Post_Content_Filters::add_filters();
 		$GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] = true;
-		echo self::render_and_restore_post_globals( (int) $_POST['post_id'], false, $panels_data );
-		SiteOrigin_Panels_Post_Content_Filters::remove_filters();
-		unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
+
+		try {
+			echo self::render_and_restore_post_globals( (int) $_POST['post_id'], false, $panels_data );
+		} finally {
+			SiteOrigin_Panels_Post_Content_Filters::remove_filters();
+			unset( $GLOBALS[ 'SITEORIGIN_PANELS_POST_CONTENT_RENDER' ] );
+		}
 
 		wp_die();
 	}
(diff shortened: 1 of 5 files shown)


Misplon:
Closing without merging. The reported case (#1370, a form shortcode baked into the Copy Content mirror by the Editor widget) is fixed by so-widgets-bundle#2364, which is on SOWB develop. This PR would extend that to every widget from every plugin, reversing how the core Text widget has behaved in the mirror since 2017. That is a behaviour change with an unmeasured blast radius (REST and headless consumers, search indexers and email plugins that read post_content raw and expect HTML) and no report asking for it, so it does not earn a release on its own.

If a report arrives for a non-Editor widget, this re-lands as-is: one commit, 14-method suspension test plus the parity exception case, verified live through the classic editor save path. Companion so-widgets-bundle#2366 is closed alongside.

Misplon:
Correction to the closing note above. The reporter's shortcode is in a **Custom HTML widget**, not the SiteOrigin Editor widget, so so-widgets-bundle#2364 does not cover their case, and the beta they were offered was this branch.

Why it never worked there: core's Custom HTML widget does not run shortcodes itself (core hooks `do_shortcode` on `widget_text_content`, which only the Text widget uses). On a live page the shortcode runs because Page Builder renders on `the_content` at priority 10 and core's `do_shortcode` follows at 11. The Copy Content mirror is rendered without `the_content` filters, so with plain core it keeps the shortcode literally and `has_shortcode()` checks on `post_content` pass. The failure needs a `widget_text` → `do_shortcode` filter on the site, which executes the shortcode during the mirror render and bakes the placeholder. That combination has behaved this way since Copy Content began in 2017; it is not a regression. The Editor widget case was (so-widgets-bundle 1.58.11), and #2364 fixes that one.

The reporter has been told the global change is not shipping, offered the Editor widget or turning Copy Content off, and asked what on their site adds shortcode handling to text widgets. Decision pending that answer:

- If it is a common source (Ninja Forms itself, say), a narrow exemption for `WP_Widget_Custom_HTML` in the mirror render is worth a separate, small PR. It would not touch the core Text widget's 2017 behaviour.
- If it is their theme or a snippet, this stays closed and the answer is the Editor widget.

Ticket in FreeScout: search `2.36.1-beta`.

PR #1374, opened 2026-09-05
Builder: Never Save a Layout That Failed to Load

Misplon:
Stacked on #1367. Refs #1366.

**Reading order.** The first two commits answer the review on #1367: `sanitize_all()` now removes a null or scalar style through one helper that uses `array_key_exists()` (the `isset()` blocks could never see a stored null), and the three save paths run that helper on the *output* of `siteorigin_panels_data_pre_save`, since `sanitize_all()` runs before that filter and the Toggle Visibility migration writes its null there. Then the ten fail-closed commits from this PR's first version, unchanged apart from one context line. Then four small follow-ups: a test-stub compatibility line, and three decoder tightenings so a submitted layout the builder could not load (numeric-keyed objects for lists, cells or widgets pointing past the rows, a row that is not an array) is refused before it replaces stored data.

**What the fail-closed commits do.** When a stored layout throws while the builder loads it, the builder no longer serializes the rows that did load. It records the failure, hands out the stored layout instead of the partial one, locks its controls behind a notice with the error and a Reload button, and disables the hidden `panels_data` field, so Update sends no layout and `save_post()` leaves the stored one untouched while title and content save normally. The Live Editor, the in-dialog Save, the nested Layout Builder widget and the Layout Block are closed off the same way. `save_post()`, `save_home_page()` and the Layout Builder widget refuse a submission that is not a layout instead of decoding it to an empty one and deleting what is stored. Nothing changes for a layout that loads cleanly.

**Why.** Reproduced on develop through the classic editor: a page with 3 widgets and 2 rows, one row style stored as null, saved as 0 widgets and 1 row after a single Update. With this stack the same Update leaves the stored layout byte-identical, and the null can no longer be written in the first place.

**Merge path.** Merge #1367 first, as-is; GitHub retargets this PR at develop; merge this next. The first two commits here are what make #1367's cleanup effective, so the two should ship in the same release. @AlexGStapleton if you'd rather carry those two commits inside #1367, cherry-pick `bbd27ce3` and `14960a26`; nothing on your branch was touched.

**Tests.** Node harness for the builder model (`nvm use 20 && npm ci && npm test`, 14 tests), PHP: 90 + 6 + 28 + 58 + 3 across the five suites, including real `save_post()` and `save_home_page()` driven with a null-writing pre-save callback. Verified in the browser on the classic editor, the nested widget, the Layout Block, an unparseable field, the admin-bar Live Editor link, Revert to Editor and the inline Save, and on the live site through the real `save_post()` with a Premium-shaped null-writing filter.


Misplon:
Merging: #1367 landed first as-is, this PR retargeted to develop. Since the 5 Sep browser pass, re-verified today on the develop-merged tree: PHP suites 90+6+3+58+28 and node 14 green; classic page with null row and widget styles keeps 3 widgets / 2 rows on Update with nulls removed; healthy page saves byte-identical to develop; nested Layout Builder widget and Layout Block with the same nulls load, save and render; a stale bundle reproduced the original crash and the builder locked with the stored layout untouched. Independent cross-model review (Codex) cleared the combined stack.

Issue #1376, opened 2026-09-14
WPML 5.0: Test Compatibility And Update Documentation


Misplon:
## Summary

WPML 5.0 is a major release. The WPML compatibility team wrote to us on 11 September 2026 (FreeScout #85127) with three code items and two documentation items. We grepped Page Builder, the Widgets Bundle and SiteOrigin Premium. No code change is needed. This issue tracks the work to do once 5.0 ships.

## Code Audit

All four greps from the WPML notice were run across `siteorigin-panels`, `so-widgets-bundle` and `siteorigin-premium`.

| Item | Grep | Result |
| --- | --- | --- |
| Admin links | `page=wpml` | No hits |
| Language switching | `wpml_switch_language`, `switch_lang` | No hits |
| Language codes | `ICL_LANGUAGE_CODE`, `wpml_current_language`, `wpml_default_language` | 6 hits, none compared against a hard-coded code |
| Legacy routers | `icl_ajx_action`, `icl_tm_action` | No hits |

Language code hits, for reference:

- `inc/sidebars-emulator.php:78` strips the language prefix from the request path. A country suffix such as `en-us` passes through.
- Widgets Bundle `compat/block-editor/widget-block.php:461` and `base/inc/actions.php:119` pass the code straight into `wpml_object_id` or the `icl_translations` query.
- Premium `custom-layouts.php:329` and `logo-booster.php:59,88` do the same.

### Test

- [ ] Install WPML 5.0 on the test site.
- [ ] Translate a Page Builder page with the Advanced Translation Editor.
- [ ] Translate a Layout Block and a Widget Block.
- [ ] Translate a page with a Layout Builder widget.
- [ ] Check the Post Loop and Custom Layouts post selectors return posts for the current language only.
- [ ] Check Logo Booster still loads the per-language logos.
- [ ] Confirm the `en-us` style codes work on a fresh site with English (United States).

### Update Documentation

- [ ] [Using Page Builder with WPML](https://siteorigin.com/page-builder/compatible-translation-plugins/using-page-builder-wpml/): replace "WPML → Translation Management" with the Translation Dashboard path, change "credits" to "words", and re-check the Language metabox steps.
- [ ] [WPML Translation Editor Page Builder Compatibility](https://siteorigin.com/wpml-translation-editor-page-builder-compatibility/): retake the two WPML screenshots and re-check the "Translate automatically" step.

### Report

[...]

PR #1378, opened 2026-09-15
Sidebars Emulator: Strip The WPML URL Prefix, Not The Language Code

Misplon:
## Summary

WPML 5.0 separates a language's code from the URL directory it is served at. A site can store German as `de` and serve it at `/de-de/`, or `ar` at `/arabic/`. The Sidebars Emulator stripped the language *code* from the request path before `get_page_by_path()`, so on such a site the prefix stayed in the path, the lookup found nothing, and the page's emulated sidebar widgets were not registered. No error was reported.

Reported by the WPML Compatibility team while reviewing our WPML 5.0 audit (#1376).

## Change

`inc/sidebars-emulator.php`, `register_widgets()`:

- Read the URL prefix from the `wpml_language_codes_map` filter, the same map WPML builds its own links from, and strip that instead of the code. WPML 4.x returns the code unchanged, so one release covers both versions.
- `preg_quote()` the prefix before interpolating it into the pattern.
- Fall back to the code when the map lacks the current language or holds an empty or non-string value.

## Tests

New isolated suite `phpunit-sidebars-emulator.xml` → `tests/sidebars-emulator/`. It runs on its own config because `tests/AbilitiesTest.php` replaces the emulator class with a spy in the default suite. Wired into `composer test`.

| Case | Request path | Reaches `get_page_by_path()` as |
| --- | --- | --- |
| Mapped prefix (`de` → `de-de`) | `/de-de/about/` | `/about/` |
| Identity map (WPML 4.x) | `/de/about/` | `/about/` |
| WPML inactive | `/de/about/` | `/de/about/`, map filter never applied |
| Prefix with a metacharacter (`en.us`) | `/en.us/about/` | `/about/` |
| Metacharacter is quoted | `/enXus/about/` | `/enXus/about/` |
| Partial map, empty value, non-string value | `/de/about/` | `/about/` |
| Prefix `"0"` | `/0/about/` | `/about/` |
| End to end, mapped prefix | `/de-de/about/` | `option_widget_text` filter registered |

Mutation checks: removing `preg_quote()`, ignoring the map, and dropping the `is_string()` guard each fail only the tests written for them.

## Verification

- `composer test`: all six suites green.
- WPML 4.7.4, real filters, real `register_widgets()`: `/pt-pt/consulting/` registers the same six `option_widget_*` filters as `/consulting/`.

## Merge Gate

Not for merge yet. The live WPML 5.0 check (German served at `/de-de/`, emulated widget renders, and the map holds `de => de-de` at `widgets_init`) gates this PR. The 5.0 build is requested from the WPML Compatibility team and not yet available. Results will be posted here before merge.

## Out Of Scope

On a subdirectory install the emulator's `^/<prefix>/` anchor and full-path `get_page_by_path()` lookup never match, with or without WPML, before and after this change. Pre-existing; not touched here.

See #1376.


Misplon:
## WPML 5.0 Live Test: Pass

Build: `sitepress-multilingual-cms.5.0.0-pre.1` from the WPML Compatibility team. Local site, WPML 4.7.4 backed up and restored afterwards (`diff -rq` identical).

### Contract In The 5.0 Source

- The filter is hooked at plugin load: `inc/functions-load.php:116` adds `WPML\Language\ActiveLanguagesReadModel::applyUrlCodeMap` before the URL converter is built.
- `applyUrlCodeMap()` rewrites each `code => code` entry to `code => display_code` where the two differ (`classes/language/ActiveLanguagesReadModel.php:49-60`).
- WPML's own canonical redirect reads it the same way we do: `apply_filters( 'wpml_language_codes_map', array( $lang => $lang ) )` guarded by `isset && is_string && '' !==` (`classes/canonicals/class-wpml-canonicals-hooks.php:99-103`).

### Live Run

Portuguese stored as `pt-pt`, `display_code` set to `portugues`.

| Check | Result |
| --- | --- |
| `apply_filters( 'wpml_language_codes_map', [ 'pt-pt' => 'pt-pt' ] )` | `[ 'pt-pt' => 'portugues' ]` |
| WPML's own permalink for the page | `/portugues/consulting/` |
| Map at `widgets_init` priority 0 on a real frontend request | `{"pt-pt":"portugues"}`, language `pt-pt` |
| `register_widgets()` on `/portugues/consulting/`, this branch | 6 `option_widget_*` filters registered |
| `register_widgets()` on `/portugues/consulting/`, `develop`'s file | 0 filters registered (the defect) |
| `register_widgets()` on `/consulting/` (English) | 6 filters, unchanged |

Restored: prefix reset, WPML back to 4.7.4 and inactive as before, probe mu-plugin removed.

The merge gate is met.

